Issue #34, opened 2026-07-01
Skill suggestion from repeated instructions


ajdelaguila:
Detect repeated instructions across the KB and propose converting them into a reusable skill or workflow. Depends on the telemetry data.

ajdelaguila:
Maintainer triage: accepted for backlog. This should become a scoped skill capture and suggestion flow rather than an automatic instruction rewrite.

Issue #33, opened 2026-07-01
Cross-agent skill distribution


ajdelaguila:
Import team-shareable skills from the MCP on first agent run, with a per-agent installer analogous to the SessionStart hook installer. Built-in and custom skills. Hardest piece: cross-agent skill formats differ.

ajdelaguila:
Maintainer triage: accepted. Cross-agent skill distribution remains useful, but it needs packaging and trust boundary design before implementation.

ajdelaguila:
The Omni Agent proof of concept validates this issue and narrows the distribution contract.

The portable package should carry:

* A package version and contract version.
* Entry point content digests.
* Supported agent runtimes and minimum CLI contracts.
* Deterministic install targets for personal and project scope.
* An on-use drift check that compares installed package version, entry point digests, CLI version, and CLI help digest.
* A fail-closed `update_required` result when the installed contract is unsupported. Do not silently update during an active task.
* A rollback target for the previous known-good package.
* A conformance suite that each agent runtime can run after installation.

The proof of concept also showed that instructions alone are insufficient for CLI safety. The distributed skill should include deterministic adapters and tests, while Data Olympus remains the version and discovery authority rather than duplicating the full manuals in the knowledge base.

Issue #239 now tracks the portable handoff, protected receipt, and execution contract that this distribution mechanism would deliver.

Issue #32, opened 2026-07-01
Consultation telemetry + per-user stats + reporting


ajdelaguila:
Aggregate kb_consult/kb_compliance events into per-principal stats for reporting. Privacy-gated: config opt-in, stored on the state volume, never committed to the public KB. Feeds curation and skill suggestion.

ajdelaguila:
Maintainer triage: accepted. Consultation telemetry and reporting are valuable, but they need privacy review and aggregate reporting design before implementation.

Issue #31, opened 2026-07-01
Curation: pattern promotion (kb_curate)


ajdelaguila:
Surface repeated patterns across the KB and propose hoisting them up the tier chain (component to project to universal) via kb_propose_edit. Advisory and human-gated. Name avoids collision with the existing kb_audit event-log tool.

caioribeiroclw-pixel:
This is a good place to keep the "agent observed a pattern" vs "team has accepted a rule" boundary sharp.

One compact shape that may help `kb_curate` stay advisory/human-gated is a **promotion receipt** alongside each proposed hoist, rather than just a ranked suggestion:

```json
{
  "candidate_id": "promote:retry-policy:2026-07-09",
  "from_tier": "T4",
  "to_tier": "T3",
  "proposed_status": "draft",
  "source_concepts": [
    {"id": "component-a-retry-policy", "status": "active", "hash": "sha256:..."},
    {"id": "component-b-retry-policy", "status": "active", "hash": "sha256:..."}
  ],
  "supporting_signals": {
    "repeated_pattern_count": 4,
    "distinct_paths_or_components": 3,
    "recent_consult_hits": 7
  },
  "negative_controls": [
    {"id": "batch-job-exception", "reason": "explicitly different latency contract"}
  ],
  "conflict_candidates": ["legacy-retry-guidance"],
  "blast_radius": {"affected_tiers": ["T3", "T4"], "estimated_concepts": 9},
  "requires_review": true,
  "stale_if": [
    "source concept hash changes",
    "source concept is superseded/deprecated/rejected",
    "conflict candidate becomes in-force",
    "bundle index rebuilt with different relationship graph"
  ]
}
```

The useful invariant: `kb_curate` should be able to say "these repeated facts justify proposing a draft at a higher tier" without making the higher-tier rule authoritative until review. That also makes false positives easier to debug: was the promotion based on real repeated in-force concepts, old superseded text, or one noisy component pattern being overgeneralized?

I would especially keep the negative-control slot. If the tool can name one near-match that should *not* be promoted, reviewers can catch the common failure mode where agents turn a local workaround into a universal standard.

ajdelaguila:
Maintainer triage: useful proposal. The promotion receipt shape matches the human-gated curation direction. Keep open for a dedicated kb_curate design pass.

ajdelaguila:
Reopening: this was closed by automation, not by a decision.

[...]

PR #35, opened 2026-07-01
feat(onboarding): guided onboarding, discovery capture & cleanup

ajdelaguila:
## Guided onboarding, discovery capture & cleanup

Adds a guided project-onboarding experience on top of the existing onboarding primitives (`kb_onboarding_status`, `kb_bootstrap_project`). MCP prompts drive an import + interview + bootstrap + cleanup flow, backed by a new read-only `kb_cleanup_plan` tool that detects repo/KB duplication and returns thin-pointer replacements.

### What's added
- `thin_pointer.py`: thin-pointer renderer (data-olympus owns the template, so every deployment emits identical pointers).
- `dedup.py`: deterministic repo/KB overlap heuristics (exact hash; heading-set + k-shingle Jaccard for partial overlap).
- `kb_cleanup_plan_fn` + `CleanupItem`/`CleanupPlanResponse`: read-only classification of local docs vs committed KB content, with thin-pointer text for exact duplicates.
- MCP tool `kb_cleanup_plan` + REST mirror `POST /api/v1/onboarding/cleanup-plan`.
- `onboarding_playbook.py`: single-sourced guided script (dispatch / project / component).
- Three MCP prompts (`onboard`, `onboard_project`, `onboard_component`) as thin wrappers over the playbook.
- Playbook `GET /api/v1/onboarding/playbook` endpoint + `kb onboard playbook` CLI + mock-server extension (cross-agent fallback for agents that do not render MCP prompts).
- `ROADMAP.md` indexing the deferred pieces B-E.

### Design invariants (test-backed)
- **Server proposes, agent applies:** the cleanup tool and route are read-only; the server never edits a project repo.
- **Single-sourced playbook:** MCP prompts, the REST/CLI endpoint, and the bats mock-server all render from `render_playbook`; no divergent copies.
- **Additive only:** existing write/bootstrap primitives are unchanged.

### Testing
- Full pytest suite green (1 unrelated pre-existing skip: `sentence_transformers`).
- bats onboarding 3/3 (includes the new `kb onboard playbook` test).
- ruff clean.

### Review
Whole-branch review flagged input-validation/compute-bound gaps on the public, unauthenticated `cleanup-plan` REST route; hardened (400-not-500 on malformed `local_files`/`jaccard_threshold`, file-count cap via `max_bootstrap_files`, defensive fn + shingle precompute) and re-reviewed clean.

### Roadmap follow-ups (tracked)
- #31 Curation / pattern promotion (`kb_curate`)
- #32 Consultation telemetry + per-user stats (privacy-gated)
- #33 Cross-agent skill distribution
- #34 Skill suggestion from repeated instructions

### Minor deferred (non-blocking)
- `kb onboard playbook` CLI prints a raw traceback on a non-2xx response instead of a clean `kb: ...` message (fallback surface).
- dedup heuristic tuning (heading-level awareness, punctuation in shingles) deferred per design; affects only the advisory partial-overlap signal, not the exact-duplicate path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ This project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Cross-agent enforcement providers: `kb enforce install --agent codex|gemini|opencode|copilot-cli|copilot-ide` and `--all`. Codex and Gemini get hard PreToolUse/BeforeTool gates (merging into existing hooks), OpenCode gets a `tool.execute.before` plugin, and Copilot CLI/IDE get a soft instructions-file + MCP provider. Antigravity is a documented-unsupported stub. The `kb-enforce-hook` dispatcher gained a `--dialect gemini` output mode.
 - Detection floor for un-hookable agents: `kb enforce report` (and `data-olympus report`) correlates governed git commits against the consult audit and lists changes with no consultation on record. An opt-in repo-scoped git provider (`kb enforce install --agent git`) installs a post-commit warning hook, or a pre-commit blocking hook with `--block`. Reuses the existing audit endpoint; no server change.
 - Enforcement hardening and observability: gate_bypass and gate_degraded events are now recorded (via `data-olympus report --emit-events` / the git warn hook, and the pre-tool hook on a reachable-degraded gate), so `kb_compliance` surfaces them. The gate receives `action_diff` and the classifier uses word-boundary matching plus dependency-install command signals (Codex and Claude now gate the Bash tool). New `kb enforce install --mode off|soft|hard`, ledger persistence via `KB_LEDGER_PATH`, a friendly PATH hint for `kb enforce report`, and a CI guard requiring a changelog entry for functional changes.
+- Guided onboarding: MCP prompts `onboard`, `onboard_project`, and `onboard_component` walk an agent through bootstrapping a new workspace or component, backed by a single-sourced playbook (`render_playbook`). A read-only `kb_cleanup_plan` MCP tool plus `POST /api/v1/onboarding/cleanup-plan` classify local repo docs against KB content and propose thin-pointer replacements for duplicates. `GET /api/v1/onboarding/playbook` and `kb onboard playbook` expose the same script to agents without native MCP prompt support.
 
 ### Fixed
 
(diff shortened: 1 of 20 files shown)


ajdelaguila:
Review findings. I attempted to submit this as a requested-changes review, but GitHub rejects request-changes reviews on your own pull request.

- P1: PR is not mergeable because `changelog-guard` is failing.
  This PR adds new CLI, MCP, and REST behavior (`bin/kb`, `server.py`, `rest_api.py`) but does not update `CHANGELOG.md`. The guard in `scripts/check_changelog.py` requires a changelog entry for functional changes, matching `.rules/changelog-per-release.md`.

- P2: `cleanup-plan` returns 500 for non-string `content`.
  The REST route validates `path` but not `content`, then `kb_cleanup_plan_fn` passes `lf.get("content", "")` into string-only dedup code. I reproduced `{"content": null}` returning `500 Internal Server Error`. Return a clean `400`, and apply equivalent validation to the MCP path.

- P2: MCP `kb_cleanup_plan` bypasses the REST caps.
  REST caps body size and file count, but the MCP tool calls directly through to `kb_cleanup_plan_fn`. Since matching is roughly `local_files * kb_docs`, the shared function or MCP wrapper should enforce the same count/size limits.

- P2: `jaccard_threshold` accepts invalid numeric values.
  `float(...)` accepts `nan`, `inf`, negatives, and values over `1`, and the MCP tool accepts the float directly. Negative thresholds can mark nearly everything as overlap; NaN makes overlap comparisons fail. Validate finite `[0.0, 1.0]`.

- P3: `kb onboard playbook` reports server errors as Python tracebacks.
  `curl` succeeds on HTTP 400, then the CLI indexes `["text"]` unconditionally. `--kind bogus` will print a traceback instead of the API error.

Verification performed:

- GitHub CI: `test` and CodeQL passed; `changelog-guard` failed.
- Local targeted tests: `36 passed`.
- `uv run ruff check .`: passed.
- `uv run --extra dev mypy src`: passed.
- Full local `pytest`: `557 passed, 1 skipped, 1 failed`; the failure is local-path related because this review worktree lives under `.worktrees`, which the lint discovery intentionally excludes. CI's normal checkout does not hit that path issue.

ajdelaguila:
Thanks for the review. All findings addressed in 5ee057d (CI now green, including `changelog-guard`).

**P1 - changelog-guard:** Added `Added` and `Fixed` entries under the existing `[Unreleased]` block in `CHANGELOG.md`.

**P2 - 500 on non-string `content`:** `{"content": null}` now returns a clean `400`. Root cause was `entry.get("content", "")` returning `None` for an explicit null, which then hit string-only dedup code.

**P2 - MCP bypasses REST caps (root cause):** Good catch. The earlier hardening lived only in the REST handler, so the MCP tool called straight through. Validation is now centralized in `kb_cleanup_plan_fn` (raises a typed `CleanupInputError`, runs before any index access). Both surfaces wrap it: REST returns `400`, MCP returns `{"status": "rejected_invalid_input", ...}`, and both pass `max_files`/`max_content_bytes` from config so the file-count and per-file-size caps apply on both paths.

**P2 - `jaccard_threshold` invalid numerics:** Now validated as finite and within `[0.0, 1.0]` via `math.isfinite` plus a range check. `nan`, `inf`, negatives, and values over `1` are rejected; `0.0` and `1.0` are accepted.

**P3 - CLI tracebacks:** `kb onboard playbook` now checks the HTTP status and prints a clean `kb: <error>` (from the API `error` field) on non-2xx, exiting non-zero. A bats test asserts no `Traceback` on `--kind bogus`.

One intentional contract change worth flagging: a `local_files` entry missing `path` is now rejected on both surfaces (previously the shared function tolerated it with an empty `local_path`). The pre-existing tolerance test was inverted to assert rejection.

Verification: pytest 579 passed / 1 pre-existing unrelated skip, bats 4/4, `ruff check` clean, `mypy src` clean.

Issue #43, opened 2026-07-01
Reliability: MCP transport/session accumulation under repeated handshakes


ajdelaguila:
## Summary
Confirm and, if needed, fix accumulation of MCP transports/sessions under
repeated handshakes.

## Motivation
Repeated MCP handshakes each create a new streamable-HTTP transport with a fresh
session id. We need to confirm these are reaped and do not accumulate over
long-lived server uptime.

## Proposed approach
Instrument the transport/session lifecycle (count live sessions); verify cleanup
on client disconnect / handshake failure; add a bound or reaper if sessions can
leak.

## Acceptance criteria
- Sustained handshake load does not grow the live session count without bound.
- A metric / log surfaces the current live session count for observability.

Issue #39, opened 2026-07-01
Search: exact-id and tag short-circuit


ajdelaguila:
## Summary
Resolve id-shaped and exact-tag queries directly instead of competing on full-text
scoring.

## Motivation
Queries that are an exact document id or an exact tag are unambiguous lookups.
They should return the target document first rather than ranking it against
full-text noise.

## Proposed approach
In the expand-query / re-rank stages, detect id-shaped queries and exact tag
matches and boost (or direct-hit) the corresponding document.

## Acceptance criteria
- Searching a known document id returns that document as the top hit.
- An exact tag query strongly favours documents carrying that tag.
- Ordinary full-text queries are unaffected.

## Depends on
Search pipeline refactor (foundational issue in this milestone).

Issue #38, opened 2026-07-01
Search: synonym & acronym query expansion


ajdelaguila:
## Summary
Expand query terms with curated synonyms and acronyms to improve recall on the
half-remembered query.

## Motivation
Governance knowledge is jargon-dense (short forms and long forms of the same
concept, common acronyms, and so on). Exact-term matching misses obvious
equivalences that a human would consider the same concept.

## Proposed approach
A curated synonym/acronym map applied in the expand-query stage: matched terms
are OR-expanded (optionally down-weighted so a synonym hit ranks below a direct
hit). The map is config-loadable with sensible built-in defaults.

## Acceptance criteria
- A short-form query matches documents that use the long form (and vice versa).
- The synonym map is configurable without a code change.
- No measurable latency regression.

## Depends on
Search pipeline refactor (foundational issue in this milestone).

Issue #37, opened 2026-07-01
Search: status-aware ranking (down-rank superseded/deprecated rules)


ajdelaguila:
## Summary
Make search ranking aware of a document's `status` so an in-force rule outranks a
retired one.

## Motivation
data-olympus governs decisions, and its core value is distinguishing a current
decision from one that was superseded. Today ranking is pure BM25 with no notion
of `status`, so a retired document can rank above the active one that replaced
it, which is the exact failure the tool exists to prevent. This is a correctness
fix as much as a relevance improvement.

## Proposed approach
Apply a configurable status multiplier in the re-rank stage: boost in-force
documents, penalise retired or draft ones, stay neutral for unknown or empty
status. Status is a free-form frontmatter string, so the map must have sane
defaults and treat unrecognised values as neutral (never drop them).

## Acceptance criteria
- A query matching both an in-force and a retired document ranks the in-force one
  higher.
- Weights are configurable (env / config), with documented defaults.
- No regression on existing relevance metrics in `benchmarks/`.

## Depends on
Search pipeline refactor (foundational issue in this milestone).

PR #50, opened 2026-07-01
Serving hardening (503 fix + bounded ledger) and composable search pipeline

### LINKED ISSUE #36: Search: refactor Index.search() into a composable pipeline
## Summary
Refactor `Index.search()` into a small, composable pipeline so the planned
search enhancements can be developed in parallel without all colliding on one
function.

## Motivation
Several roadmap items (status-aware ranking, synonym expansion, corpus
co-occurrence expansion, trigram fuzzy matching, local-embedding hybrid) each
need to modify the query-building and/or ranking logic that currently lives in a
single `Index.search()` method. Editing them in parallel would produce
mutually-conflicting changes. A staged pipeline turns each enhancement into a
pluggable unit.

## Proposed shape
Split the current monolithic path into ordered, individually-testable stages:

- **expand-query**: term extraction, quoting, and (later) synonym / co-occurrence
  expansion. Produces the FTS `MATCH` expression and any auxiliary signals.
- **match**: run the FTS query (and, later, alternative match paths such as
  trigram or vector similarity).
- **re-rank**: apply score adjustments (BM25 base, plus later status/tier priors,
  hybrid blending). Produces the final ordered list of hits.

Each stage should be a plain function/hook with a stable signature so a feature
can add behaviour without touching the others.

## Acceptance criteria
- `Index.search()` behaviour is unchanged: the full existing test suite passes.
- Query-building and ranking are exposed as separate, unit-testable stages.
- No performance regression on the existing benchmark harness (`benchmarks/`).

## Note
Prerequisite for the search-enhancement issues in this milestone.

ajdelaguila:
Foundation branch for the search + reliability work (wave 1). Base for the
stacked feature PRs (#37, #38, #39, #43).

## What's here
**Serving reliability — eliminate the intermittent 503s.** Root cause: the
single-replica pod flapped NotReady under load because the readiness probe
(`timeoutSeconds: 1`) timed out while synchronous work ran on the one asyncio
event loop; TCP-only liveness meant no restart and no log trace (nginx served the
503 with an empty upstream). Fixes: relax the readiness probe (5s timeout, 5
failures) + raise CPU to 2; offload synchronous REST/enforce handler work to the
anyio threadpool; cache `Index.health()` (short TTL, invalidated on rebuild);
lock `AuditLog` and `ConsultationLedger` for concurrent thread access.

**Bounded consultation ledger.** TTL eviction + hard max-entries cap so the
ledger stops growing without bound and the per-record rewrite stays O(active
window).

**Composable search pipeline (Closes #36).** `Index.search()` refactored into
expand-query / match / re-rank stages with pluggable `query_expander` and
`reranker` hooks. Behaviour unchanged.

## Review notes
- This branch was cut before #35 merged, so it is 1 commit behind `main`. It
  needs a rebase onto `main` before merge (small conflict surface: rest_api.py,
  server.py).
- `index.py` bundles the health-snapshot cache with the pipeline refactor
  because they touch the same file; can be split on request.

Closes #36

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,9 +32,28 @@ This project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Detection floor for un-hookable agents: `kb enforce report` (and `data-olympus report`) correlates governed git commits against the consult audit and lists changes with no consultation on record. An opt-in repo-scoped git provider (`kb enforce install --agent git`) installs a post-commit warning hook, or a pre-commit blocking hook with `--block`. Reuses the existing audit endpoint; no server change.
 - Enforcement hardening and observability: gate_bypass and gate_degraded events are now recorded (via `data-olympus report --emit-events` / the git warn hook, and the pre-tool hook on a reachable-degraded gate), so `kb_compliance` surfaces them. The gate receives `action_diff` and the classifier uses word-boundary matching plus dependency-install command signals (Codex and Claude now gate the Bash tool). New `kb enforce install --mode off|soft|hard`, ledger persistence via `KB_LEDGER_PATH`, a friendly PATH hint for `kb enforce report`, and a CI guard requiring a changelog entry for functional changes.
 - Guided onboarding: MCP prompts `onboard`, `onboard_project`, and `onboard_component` walk an agent through bootstrapping a new workspace or component, backed by a single-sourced playbook (`render_playbook`). A read-only `kb_cleanup_plan` MCP tool plus `POST /api/v1/onboarding/cleanup-plan` classify local repo docs against KB content and propose thin-pointer replacements for duplicates. `GET /api/v1/onboarding/playbook` and `kb onboard playbook` expose the same script to agents without native MCP prompt support.
+- Composable search pipeline: `Index.search()` now runs expand-query / match /
+  re-rank stages with pluggable `query_expander` and `reranker` hooks, so ranking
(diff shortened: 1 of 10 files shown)


ajdelaguila:

Blocking issues:

1. `CHANGELOG.md` is not updated for this functional PR. CI already fails `changelog-guard` with: `CHANGELOG.md must be updated for functional changes`. Add an `[Unreleased]` entry covering the serving hardening, bounded ledger, and search pipeline change.

2. `src/data_olympus/rest_api.py:236` still routes `/api/v1/health` through `_offload()`, which uses the shared anyio worker-thread pool. Under worker-pool saturation from slow blocking handlers, readiness can still wait behind the workload it needs to survive, so the 503 fix is not complete. Keep health off the shared limiter, or isolate long handlers behind a separate limiter, and add a saturation regression test.

3. `src/data_olympus/enforce_policy.py:147` loads a persisted ledger without applying TTL or the hard `max_entries` cap. Upgrading from the previous unbounded ledger can still load an oversized `/state/ledger.json` into memory before the first new `record()` prunes it. Please bound/prune on load and add a persisted-ledger test.

Checked:
- `gh pr checks 50`: test, CodeQL, and Analyze pass; `changelog-guard` fails.
- CI test job includes ruff, mypy, pytest, bats, and example-bundle lint.
- `git diff --check` clean.
- No obvious secret patterns found in changed files.

PR #51, opened 2026-07-01
Search: status-aware ranking

### LINKED ISSUE #37: Search: status-aware ranking (down-rank superseded/deprecated rules)

ajdelaguila:
Status-aware search ranking. Stacked on the foundation branch (`claude/cranky-matsumoto-2e60a7`).

A governance KB must not rank a retired decision above the one that replaced it.
Adds a `reranker` (`make_status_reranker`) wired on by default: boosts in-force
statuses (active, accepted), penalises retired ones (superseded, deprecated,
draft), neutral for unknown/empty (never drops a hit). Correctly handles the bm25
direction (lower score = better) with explicit ordering tests. Configurable via
`KB_STATUS_WEIGHTS` (JSON; fails loud on malformed).

## Integration note
Composes with #39: both use the single `reranker` slot, to be combined as
`make_id_tag_reranker(idx, inner=make_status_reranker(...))` at merge.

Closes #37

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,14 @@ This project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Retrieval now indexes `applies_when` trigger metadata and `description` with
   column-weighted ranking, improving coding-intent to governing-rule matching
   (index schema v5). `kb_get` returns both fields.
+- Status-aware search ranking (enabled by default): `kb_search` reranks hits by a
+  status prior on top of BM25, so an in-force doc (`active`, `accepted`,
+  `approved`) outranks a `superseded`/`deprecated`/`rejected` one that matches the
+  same query, and drafts do not beat live guidance. Status matching is
+  case-insensitive; unknown or empty statuses stay neutral and are never dropped.
+  This changes result ordering for queries that hit both a live and a retired doc.
+  A deployment overrides the built-in status->weight map via `KB_STATUS_WEIGHTS`
+  (a JSON object of `{status: weight}`, negative boosts, positive penalizes).
 - Enforcement core: data-olympus can now act as a gated, mandatory consultation
   proxy for code/architectural decisions, not only an advisory KB. New MCP tools
   `kb_consult`, `kb_gate_check`, `kb_compliance` and REST endpoints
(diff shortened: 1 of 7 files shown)


ajdelaguila:

Core logic is clean and well-tested, but this is not merge-ready.

Blocking issue:
- `changelog-guard` is failing. This is a functional change: the status reranker is enabled by default in `build_app`, changes user-visible search ordering, and adds `KB_STATUS_WEIGHTS`. Add a `CHANGELOG.md` entry under `## [Unreleased]`.

Recommendations before merge:
- `_DEFAULT_STATUS_WEIGHTS` omits `approved`. The target KB uses `status: approved` as an in-force status, so those docs receive neutral weighting. Add `approved` with the same boost as `accepted`.
- The reranker runs after the SQL `LIMIT`, so it cannot promote an active document that falls just outside the initial BM25 window. Either rerank a wider candidate pool before truncating, or document the limitation.
- Weight deltas are fixed additive constants against BM25 scores whose scale varies by query and column weights. Acceptable for v1, but tests currently pin direction more than real-world adequacy.
- Status lookup is case-sensitive. Casefold configured status keys and document statuses to avoid silent neutral treatment for mixed-case frontmatter.
- Document `KB_STATUS_WEIGHTS` and the default-on ordering change in `docs/serving.md`.

Merge coordination:
- This PR and the exact-id/tag shortcut PR both occupy the single `reranker` slot. When merging the stack, compose them so one does not silently disable the other.

Checked: CI `test` passes; `changelog-guard` fails. New tests cover ranking direction, override config, default shape, neutral cases, and end-to-end acceptance. No secrets or prose issues found.

PR #52, opened 2026-07-01
Search: synonym & acronym query expansion

### LINKED ISSUE #38: Search: synonym & acronym query expansion

ajdelaguila:
Synonym & acronym query expansion. Stacked on the foundation branch (`claude/cranky-matsumoto-2e60a7`).

Adds a `query_expander` (curated bidirectional map: k8s/kubernetes,
auth/authentication, rls, adr, ...), enabled by default. Originals kept first so
BM25 still favours typed terms; bounded at 32 terms. Configurable via
`KB_SYNONYMS_MODE` (merge/replace/off) and `KB_SYNONYMS`. Expanded terms still
pass through `_build_match_expr` quoting, so no FTS-operator injection.

Closes #38

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,15 @@ This project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   (default 60s). `kb_health` and `/api/v1/health` gained a `live_sessions` field
   reporting the current live transport count (`null` until the HTTP app is
   serving); a value that only climbs signals leaked sessions.
+- Synonym / acronym query expansion (default-on). Before the FTS MATCH is built,
+  the search pipeline rewrites the query term list through a curated, bidirectional
+  synonym map, so a short-form query (`k8s`, `rls`) also reaches documents that only
+  use the long form (`kubernetes`, `row level security`) and vice versa. Adjacent
+  query tokens are scanned as n-grams so multi-word canonical keys match from a
+  long-form query. Expansion is bounded (32 terms) and de-duplicated, originals are
+  ranked first, and it is configurable via `KB_SYNONYMS` (extra/override groups) and
+  `KB_SYNONYMS_MODE` (`merge` default / `replace` / `off`). This is curated
+  lexical expansion, not semantic (embedding) retrieval.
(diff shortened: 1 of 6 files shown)


ajdelaguila:

The single-token synonym implementation is clean and well-tested, but there are merge blockers.

Blocking issues:
1. `changelog-guard` is failing. This PR changes `src/data_olympus/query_expansion.py` and wires expansion on by default in `server.py`, so it is a functional behavior change. Add a `CHANGELOG.md` entry under `## [Unreleased]`.
2. Multi-word synonyms only expand one direction. `Index.search` uses `query.split()`, and the expander looks up each token individually. Multi-word canonical keys like `row level security`, `pull request`, `continuous integration`, and `architecture decision record` can never match from a long-form query, so long-form to acronym expansion silently fails. Fix with adjacent-token n-gram lookup, or narrow the bidirectional claim and default map to entries that actually work. Add a test for the long-form direction.
3. Docs still contradict the new behavior. `docs/comparison.md` says search cannot follow synonyms and synonym-dependent queries miss results. Update it to describe curated synonym/acronym expansion while still distinguishing it from semantic retrieval.
4. New operator env vars `KB_SYNONYMS` and `KB_SYNONYMS_MODE` are not documented in `docs/serving.md`. Add format and mode semantics.

Nit:
- Default-on expansion for ambiguous short tokens like `cd`, `ci`, `pr`, `db`, `docs`, `config`, and `repo` can add noise. The current ordering and BM25 behavior mitigate it, but the default set should be intentional.

Checked: CI `test` passes; `changelog-guard` fails. Existing tests cover single-token bidirectionality, order, case-insensitivity, deduping, bounds, env modes, server wiring, and negative no-expander behavior. No secrets found.

PR #53, opened 2026-07-01
Search: exact-id and tag short-circuit

### LINKED ISSUE #39: Search: exact-id and tag short-circuit

ajdelaguila:
Exact-id and tag short-circuit. Stacked on the foundation branch (`claude/cranky-matsumoto-2e60a7`).

Detects id-shaped queries (single token, id pattern) gated on a real
`Index.get`, and exact tag matches, and puts the target document first — even
when SQL's `LIMIT` would exclude it (prepends via lookup). New
`search_shortcut.py` + `Index.ids_with_exact_tag`. Ordinary multi-term queries
are unaffected.

## Integration note
Built composably: `make_id_tag_reranker(index, inner=...)` — plug #37's status
reranker in as `inner` at merge so exact-match wins on top of status ranking.

Closes #39

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@ This project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- Search now short-circuits exact-id and exact-tag queries: a single-token query
+  that is a document id (e.g. `STD-U-002`, or a path-derived id) is surfaced as
+  the top hit via a direct lookup even when it is absent from the bm25 results,
+  and a single-token exact-tag query lifts docs carrying that tag ahead of the
+  rest. Implemented as a composable `reranker` (issue #39); the match stage and
+  `_build_match_expr` are untouched.
 - `kb_search` and `kb_get` now support and return `status` and `type`, making the
   documented "filter by status / tier / type" capability real (index schema v4).
 - Retrieval now indexes `applies_when` trigger metadata and `description` with
--- a/src/data_olympus/index.py
+++ b/src/data_olympus/index.py
@@ -651,6 +651,27 @@ class Index:
             description=row["description"] or "",
         )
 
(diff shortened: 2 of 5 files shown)


ajdelaguila:
## Review decision: recommendations

This is close, but I would address one non-blocking contract issue before merge.

Recommendation:
- The exact-id path can return `limit + 1` results. When the id document is absent from the initial FTS top-N, the reranker prepends the synthetic exact-id hit and keeps the original hit list, so `search(limit=N)` can return N+1 rows. The reviewer reproduced `search("STD-Z-999", limit=3)` returning 4 hits. Suggest truncating back to the requested limit, or documenting the intentional +1 behavior in the docstring and changelog.

Nits:
- The id and tag behaviors are asymmetric: exact-id can surface a document absent from FTS via `Index.get`, while tag shortcut only reorders documents already present in the FTS hit list. This is documented, so it is a product choice rather than a bug.
- `ids_with_exact_tag` uses a broad LIKE pattern without escaping `%` or `_`. Results remain correct because membership is rechecked after splitting tags, and the query is parameterized, so this is only an efficiency note.

Checked: CI `test` and `changelog-guard` pass. The new tests are strong, especially the FTS-absent exact-id precondition. No new em-dashes, no secrets, and the changelog entry is present.

PR #54, opened 2026-07-01
Reliability: bound MCP session growth (idle reaper + observability)

### LINKED ISSUE #43: Reliability: MCP transport/session accumulation under repeated handshakes

ajdelaguila:
MCP transport/session lifecycle — fix the session leak. Stacked on the
foundation branch (`claude/cranky-matsumoto-2e60a7`).

**Finding: sessions genuinely leak.** FastMCP 3.4.2 builds the SDK session
manager with no `session_idle_timeout` (defaults None), so a `POST /mcp`
handshake with no follow-up `DELETE` leaves the transport + its task resident
forever — unbounded growth over uptime. Cited to file:line in the installed
mcp/fastmcp packages.

Fix: `session_metrics.py` (live-session count + activity-tracking middleware +
idle reaper loop); `live_sessions` surfaced on `/api/v1/health`; knobs
`KB_SESSION_IDLE_TIMEOUT_SEC` (default 1800) / `KB_SESSION_REAP_INTERVAL_SEC`.
Includes a 50-handshake regression test asserting bounded growth.

Closes #43

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,13 @@ This project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   (`KB_ENFORCE_FAIL_MODE`), consultation freshness via `KB_CONSULT_TTL_SEC`.
 - Cross-agent enforcement providers: `kb enforce install --agent codex|gemini|opencode|copilot-cli|copilot-ide` and `--all`. Codex and Gemini get hard PreToolUse/BeforeTool gates (merging into existing hooks), OpenCode gets a `tool.execute.before` plugin, and Copilot CLI/IDE get a soft instructions-file + MCP provider. Antigravity is a documented-unsupported stub. The `kb-enforce-hook` dispatcher gained a `--dialect gemini` output mode.
 - Detection floor for un-hookable agents: `kb enforce report` (and `data-olympus report`) correlates governed git commits against the consult audit and lists changes with no consultation on record. An opt-in repo-scoped git provider (`kb enforce install --agent git`) installs a post-commit warning hook, or a pre-commit blocking hook with `--block`. Reuses the existing audit endpoint; no server change.
+- Streamable-http session idle reaper bounds transport accumulation (issue #43).
+  A background loop terminates sessions with no request activity for longer than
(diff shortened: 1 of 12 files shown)


ajdelaguila:

Solid reliability direction, but this is not merge-ready yet.

Blocking issue:
- `changelog-guard` is failing. New env knobs (`KB_SESSION_IDLE_TIMEOUT_SEC`, `KB_SESSION_REAP_INTERVAL_SEC`), the `live_sessions` health field, and the reaper task are functional changes. Add a `CHANGELOG.md` entry under `## [Unreleased]`.

Concerns:
- Quiet long-lived SSE clients can be reaped. Activity is stamped once per request in `SessionActivityMiddleware.__call__`; a streamable-http client holding an open GET SSE stream and making no periodic POST requests receives no further stamps and can be reaped after the idle window. This may be acceptable for this MCP, but `docs/serving.md` should describe the actual guarantee, or the implementation should exclude active streams.
- `SessionActivityMiddleware` lacks a direct unit test. It is the sole activity source, so regressions in header casing, header decoding, or early break behavior would either reintroduce the leak or reap live sessions early. Add an ASGI scope test with and without `mcp-session-id`.
- `session_metrics.py` docstring says `touch` is called at session creation. It is not. The real behavior comes from `idle_session_ids` stamping never-seen live ids on the first pass. Fix the docstring.

Nits:
- `session_reaper_loop` logs live count at INFO every interval, about 1440 lines per day at the default. Consider debug or log-on-change.
- `main()` replaces FastMCP `run_async` with a hand-built `uvicorn.Server`; parity with the previous serving path is low risk but not covered end-to-end.

Checked: CI `test` passes; `changelog-guard` fails. The reaper tests and health threading coverage are otherwise good. No secrets found.

PR #55, opened 2026-07-01
Scaling: read-only replica serving mode

### LINKED ISSUE #44: Scaling: read-only replica serving mode for horizontal read scale
## Summary
Provide a read-only replica serving mode so read traffic can scale horizontally,
per the single-writer / many-reader model in the spec.

## Motivation
The spec permits horizontal read replicas: the write-enabled server stays a
single owner of git state, while read-only mirrors serving the read tools may run
as many instances as needed. A default deployment currently runs one
write-enabled instance for all traffic.

## Proposed approach
A read-only serving mode that periodically pulls from the git remote and holds its
own index snapshot, with the write pipeline disabled; deployment manifests for N
read replicas alongside the single writer.

## Acceptance criteria
- Read replicas serve read queries independently, each from its own snapshot.
- Writes remain single-owner; replicas never attempt the write pipeline.
- Documented deployment topology.

ajdelaguila:
Read-only replica serving mode. Based on `main` (this branch was built on
current main, which is why it targets main rather than the foundation branch).

`KB_READ_ONLY` drops the write/enforcement-write tools and REST routes, skips
the write pipeline and push/pending loops, and keeps `git_pull_loop` + the read
tools so a replica refreshes its own snapshot from the git remote. Default
(unset) behaviour is unchanged. Adds `deploy/k8s/read-replica/` (a Deployment,
3 replicas, own Service + NetworkPolicy).

## Integration note
This overlaps the foundation PR's rest_api.py/server.py changes; the two need
reconciling once the foundation lands. Design point for review: it treats the
enforcement plane (consult/gate) as write-side and disables it on replicas, so
replicas shed search load but not enforcement load.

Closes #44

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ This project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   and query-expansion features compose without rewriting the core query. When a
   reranker is installed the query is matched against a wider BM25 candidate pool
   (over-fetched) and the reranked result is truncated back to the requested limit.
+- Read-only replica serving mode (`KB_READ_ONLY`, issue #44) for horizontal read scaling. When set, the server registers only the read tools and read REST routes; the write/enforcement-write tools and routes (propose/resolve/bootstrap/consult/gate/record-event plus the observability GETs `/api/v1/pending`, `/api/v1/audit`, `/api/v1/audit/verify`, `/api/v1/compliance`) are not registered and return 404, and the write pipeline (worktrees/push queue/pending) is never initialised, so a replica is never a git writer. A new `deploy/k8s/read-replica/` overlay (Deployment, Service, ConfigMap, NetworkPolicy, kustomization) runs N interchangeable read pods with per-pod ephemeral clone + index scratch and references a separate read-only git deploy key Secret (`data-olympus-mcp-readonly`). The replica logs an unambiguous "starting in READ-ONLY mode; write pipeline disabled" line at startup.
 
 ### Fixed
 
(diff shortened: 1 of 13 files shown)


ajdelaguila:

The read/write surface split is well-scoped and the app-layer gating looks correct, but there are merge blockers.

Blocking issues:
1. `changelog-guard` is failing. The PR touches `src/` and `deploy/` and changes runtime behavior, so add an entry under `## [Unreleased]` in `CHANGELOG.md`.
2. `deploy/k8s/read-replica/deployment.yaml` pins `ghcr.io/knaisoma/data-olympus:v0.1.1`, which predates `KB_READ_ONLY`. That image ignores the flag, so a replica launched from it can initialize the write pipeline and become a second git writer using the write-capable deploy key. Pin to an image or digest built from this branch, and consider failing fast if read-only mode is requested but unsupported.

Concerns:
- Read replicas reuse the writer `data-olympus-mcp` Secret, which is a write-capable deploy key. The app-layer boundary is right, but a compromised read pod can still push. The default should use a read-only key, not just a comment suggesting one.
- `test_read_only_write_routes_absent` does not assert 404 for observability GET routes that moved behind the read-only gate: `/api/v1/pending`, `/api/v1/audit`, `/api/v1/audit/verify`, and `/api/v1/compliance`. Add them so the route split is fully pinned.
- The PR body notes overlap with the foundation PR in `rest_api.py` and `server.py`; resolve that before merge.

Verified correct: write pipeline and audit log are gated on `kb_remote_url and not read_only`; `git_pull_loop` still runs; push and pending loops stay gated; docs describe the high-level behavior; deployment hardening mirrors the writer. CI `test` passes; `changelog-guard` fails.

PR #58, opened 2026-07-01
feat(search): optional local-embedding hybrid ranking

### LINKED ISSUE #42: Search: optional local-embedding hybrid ranking
## Summary
Optional local-embedding hybrid ranking for real paraphrase handling, with no
external API.

## Motivation
Lexical search cannot match paraphrases that share no tokens. At the current
corpus scale, real embeddings are cheap: a small local model embeds the whole
corpus in seconds at build time, and brute-force cosine over the resulting
vectors is sub-millisecond.

## Proposed approach
Embed documents at build time with a small local model; store vectors alongside
the index; hybrid re-rank by blending normalised BM25 with cosine similarity.
Feature-flagged and off by default so the zero-dependency lexical mode remains
the default.

## Acceptance criteria
- With the feature enabled, hybrid ranking beats BM25-only on a semantic
  benchmark.
- Disabled by default; no external network calls; model runs locally.
- Build-time and memory cost documented.

## Depends on

ajdelaguila:
Closes #42.

Optional local-embedding hybrid ranking. **Off by default**; the zero-dependency lexical product is unchanged and does not import any embedding code unless enabled.

## Dependency footprint (this is why it is an optional extra)
New `[project.optional-dependencies] embeddings = ["fastembed>=0.8,<0.9"]`. fastembed bundles a quantised ONNX MiniLM-class model runner on `onnxruntime` — **no torch, no external API**. Transitive additions (in `uv.lock`, dev-extra so CI can exercise it): onnxruntime, flatbuffers, protobuf, loguru, mmh3, pillow, py-rust-stemmers. Model (~small MiniLM, e.g. `BAAI/bge-small-en-v1.5`) is fetched once at enable time and cached; never at query time.

## Behaviour
- `KB_EMBEDDINGS_MODE=on` (default off): `Index.build()` embeds each doc and stores vectors in a new table (schema v8), written into the same tmp DB and swapped atomically with the FTS index.
- A hybrid reranker embeds the query once and blends normalised BM25 with cosine similarity. It composes **under** the exact-id/tag short-circuit and **after** the status prior, so an exact id/tag still wins and active docs keep their boost.
- Guarded imports: with the extra uninstalled and the flag off, nothing imports the embedding stack. Enabling the flag without the extra/model **fails loudly at startup**.
- Config: `KB_EMBEDDINGS_MODE`, `KB_EMBEDDINGS_WEIGHT` (blend, default 0.35), `KB_EMBEDDINGS_MODEL`.

## Verification
Full suite green locally except the known `test_lint` worktree artifact (passes in CI's clean checkout). Embedding integration tests skip when the model/dep is unavailable in the sandbox; CI (with network) exercises them. ruff + mypy clean.

### DIFF
--- a/src/data_olympus/index.py
+++ b/src/data_olympus/index.py
@@ -25,9 +25,10 @@ from data_olympus.cooccurrence import (
 from data_olympus.embeddings import (
     EMBEDDINGS_SCHEMA,
     Embedder,
+    EmbeddingsConfig,
+    build_embedder,
+    cosine,
     deserialize_vector,
-    embeddings_config,
-    embeddings_enabled,
     make_hybrid_reranker,
     serialize_vector,
 )
@@ -104,6 +105,18 @@ _RERANK_MAX_POOL = 200
 # a doc carries its topical signal; bounding it keeps a very long doc from
 # overrunning the model's context window and keeps build cost predictable.
 _EMBED_TEXT_MAX_CHARS = 4000
+
+# Dense (semantic) candidate SOURCE defaults (issue #42, reviewer concern 1).
+# When embeddings are enabled, search() unions the top-N docs by query-doc cosine
+# into the FTS candidate pool so a paraphrase with ZERO lexical overlap can still
+# retrieve its doc (bm25 alone never surfaces it). Two knobs, both config-driven:
+#   * DEFAULT_DENSE_CANDIDATE_COUNT: how many nearest neighbours to consider.
+#   * DEFAULT_DENSE_MIN_COSINE: a minimum cosine a neighbour must clear to be
+#     added. This is the abstention guard: a negative / out-of-scope query whose
+#     nearest neighbour is only weakly similar pulls in NOTHING, so the dense
+#     source does not blow up the false-positive rate on the negative stratum.
+DEFAULT_DENSE_CANDIDATE_COUNT = 10
+DEFAULT_DENSE_MIN_COSINE = 0.5
 
 # Status priors for the status-aware reranker (issue #37). search() orders by
 # bm25 ASCENDING (lower = better), so these deltas are ADDED to the raw score:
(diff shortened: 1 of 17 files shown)


src/data_olympus/index.py
```diff
@@ -802,6 +850,50 @@ def cooccurrence_expander(
...
+        """Return a hybrid reranker bound to this index's stored vectors.
+
+        Blends normalised bm25 with query-doc cosine (issue #42). ``embedder``
+        embeds the query once per search; ``weight`` in [0, 1] is the cosine
+        fraction of the blended score. The query is embedded lazily (only when a
+        search actually runs), and a query the model cannot embed degrades to
+        bm25 ordering. Compose this as the ``inner`` reranker under the id/tag
+        short-circuit so an exact id/tag still wins (see server.build_app).
+        """
+        return make_hybrid_reranker(
```
ajdelaguila, line 1094: This hybrid reranker only reorders the hits handed to it by `Index.search()`. The public search path still builds that candidate pool exclusively from `fts MATCH`, so a tokenless paraphrase query never reaches this code with the relevant document in `hits`. I reproduced this at the current head with embeddings enabled and vectors built: `Index.search("car")` returned `[]` for a corpus containing `automobile` but no `car`. The current acceptance test drives this reranker over a hand-built all-doc pool, so it proves the cosine math but not the shipped search behavior. The implementation needs a semantic candidate source/fallback for zero or low lexical-hit queries, or the PR should not claim tokenless paraphrase retrieval through public search.

src/data_olympus/server.py
```diff
@@ -269,9 +275,32 @@ def build_app(
...
+    # candidate pool; it is composed as the ``inner`` of the id/tag short-circuit
+    # so an exact id/tag still wins, and it wraps the status reranker so an active
+    # doc still outranks the superseded one among semantically-blended hits. When
+    # disabled the stack is exactly today's (status inner, id/tag outer). The
+    # embedder is loaded once here (loud failure if enabled but unavailable).
+    status_reranker = make_status_reranker(config.status_weights)
+    inner_reranker = status_reranker
+    if config.embeddings_enabled:
+        from data_olympus.embeddings import build_embedder, embeddings_config
+        embedder = build_embedder(embeddings_config())
```
ajdelaguila: `build_app()` and `build_app_from_config()` thread `embeddings_model` into `Config`, but the enabled path ignores that value here by calling `embeddings_config()`, which rereads `KB_EMBEDDINGS_MODEL` from the environment. I reproduced the mismatch by passing `embeddings_model="param-model"` while `KB_EMBEDDINGS_MODEL="env-model"`; the loaded model was `env-model`. The same env-vs-Config split exists during index build (`Index.build()` checks `embeddings_enabled()` from env), so programmatic Config callers can report one embeddings config while building/querying with another. Build the `EmbeddingsConfig` from `config.embeddings_model` / `config.embeddings_weight`, and make the build path use the same resolved config instead of rereading env.

ajdelaguila (commented):
I re-reviewed the completed implementation at `3b38482`. The optional dependency wiring and vector storage are in place, and CI is green, but there are still two behavioral issues to fix before this satisfies issue #42 and the Config contract.

src/data_olympus/index.py
ajdelaguila: Fixed in b510dd5. Added a dense candidate SOURCE (`Index.dense_candidates` + `_union_dense_candidates`, wired into `search()` before the reranker) so a tokenless query retrieves via cosine over `doc_vectors`, min-cosine-guarded (default 0.5) so negatives do not gain false positives; with embeddings off `search()` stays byte-for-byte pure FTS. The acceptance test now drives the public path (`Index.search("car")` returns the `automobile` doc). Honest caveat: the 18-doc governance benchmark cannot show a delta because the wide FTS over-fetch already returns every doc there; the dense source pays off on genuinely tokenless queries and larger corpora.

src/data_olympus/server.py
ajdelaguila: Fixed in b510dd5. `Index` now takes an `EmbeddingsConfig` + embedder; `build()`/`search()` use those instead of rereading env. `build_app` resolves the `EmbeddingsConfig` once from `config.embeddings_model`/`embeddings_weight` and threads it (embedder built once, reused by the reranker); `load_config()` is the only env reader. Added a test asserting `build_app(embeddings_model=X)` is honored even when `KB_EMBEDDINGS_MODEL` differs.

ajdelaguila (commented):
Rechecked the fixes at `b510dd5`. The two previous concerns are resolved: public `Index.search()` now unions dense candidates before hybrid reranking, so tokenless paraphrase retrieval works through the shipped path, and embeddings config is now threaded from `Config` into `Index`/the shared embedder instead of rereading `KB_EMBEDDINGS_*` on the enabled path.

Verification: targeted embeddings tests pass (`48 passed`), the prior `Index.search("car")` reproduction now returns `DOC-AUTO`, the prior `embeddings_model` env-vs-param reproduction now loads the param model, `ruff check .` passes, `mypy src` passes, and GitHub checks are green. Full pytest has only the known `.worktrees` path artifact in `tests/test_lint.py`. No remaining blocker from my side.

PR #59, opened 2026-07-01
feat(release): automated release-cutter routine, SemVer tooling, and tag-on-bump CI

ajdelaguila:
## Summary

Adds a scheduled, human-gated release pipeline for data-olympus, conformant with STD-U-810. A daily `data-olympus-release-cutter` routine (05:00 UTC) detects a releasable `main`, runs quality gates, computes the SemVer bump from Conventional Commits, and opens a release PR with human-friendly notes. Merging that PR cuts an annotated `vX.Y.Z` tag and GitHub Release via CI, which builds the image. No tags are ever cut unattended.

## What is included

- **Bump engine** (`scripts/compute_release.py`): pre-1.0 mapping (`feat!:`/`BREAKING CHANGE:` -> minor; `feat:` -> patch; `fix:`/`perf:` -> patch; other types -> none, floored to patch only when a functional path `src/`, `bin/`, `deploy/`, `SPEC.md` changed). Single source of the parser grammar, reused by the two scripts below.
- **Tag decision** (`scripts/should_tag.py`): emits `vX.Y.Z` only when `pyproject.toml`'s `[project]` version has no matching tag.
- **PR-title lint** (`scripts/lint_pr_title.py` + `.github/workflows/pr-title-lint.yml`): the commitlint equivalent (no Node here); reuses the bump parser so the accepted grammar matches exactly. PR title flows via an env var, no shell interpolation.
- **Preview version** (`scripts/compute-version.sh`): STD-U-810 section 5 version string plus a `check` mode enforcing the section 5.1 invariant.
- **CI**: `tag-release.yml` cuts the annotated tag + Release on a version bump (idempotent, re-runnable), then builds the image via a new reusable `release-image-reusable.yml` (adds OCI version/revision labels). `release-image.yml` now calls the reusable workflow and keeps its tag trigger as the STD-U-810 section 9 manual-tag fallback.
- **Docs/config**: `.rules/versioning.md`, `.rules/release-routine.md`, `docs/releases/README.md`, a `CONTRIBUTING.md` Releases section, and a `pyproject.toml` version drift fix (`0.1.0` -> `0.1.1`, matching the latest tag).

## Design notes

- Human gate is the operator merging the release PR; the routine never runs `git tag` and never publishes externally (the social announcement is a draft in the PR body only).
- `pyproject.toml` is the single version source of truth. The tag-push uses the built-in `GITHUB_TOKEN` on purpose, so `release-image.yml` does not double-fire (documented inline in `tag-release.yml`).
- Scripts run both under pytest and as direct scripts (a `sys.path` shim, locked by subprocess regression tests).

## Follow-ups (not in this PR)

- Provision the live cron routine and run the first dry-run (expected first release `v0.1.2`). Post-merge, operator-gated.
- Confirm the repo allows `github-actions[bot]` to push `v*` tags.

## Type of change

- [x] Tool / code change (CLI, MCP server, `kb` client, deploy)
- [ ] Spec / format change (requires a linked Spec Proposal issue with maintainer sign-off)
- [x] Documentation
- [x] CI

## Checklist

- [x] Tests pass (`uv run pytest`)
- [x] `ruff` reports no errors (`uv run ruff check .`)
- [x] `data-olympus lint` exits 0 on any bundle touched by this PR (no bundle touched)
- [x] Documentation updated (if the changed behaviour is documented)
- [x] `SPEC.md` and the validator are consistent (required if the schema changed) (schema unchanged)
- [x] No em-dashes in any prose added or modified by this PR
- [x] No credentials or secrets committed

### DIFF
--- a/.github/workflows/pr-title-lint.yml
+++ b/.github/workflows/pr-title-lint.yml
@@ -1,0 +1,17 @@
+name: pr-title-lint
+on:
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
+permissions:
+  contents: read
+jobs:
+  lint-title:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: astral-sh/setup-uv@v7
+      - run: uv venv --python 3.13
+      - name: Validate PR title is a Conventional Commit
+        env:
+          PR_TITLE: ${{ github.event.pull_request.title }}
+        run: uv run python scripts/lint_pr_title.py "$PR_TITLE"
--- a/.github/workflows/release-image-reusable.yml
+++ b/.github/workflows/release-image-reusable.yml
@@ -1,0 +1,62 @@
+# .github/workflows/release-image-reusable.yml
+name: build-and-push-image
+on:
+  workflow_call:
+    inputs:
+      tag:
+        description: "image tag to publish, e.g. v0.1.0 or edge"
+        required: true
+        type: string
+      move_latest:
+        description: "also move the :latest channel"
+        required: false
+        default: false
+        type: boolean
+jobs:
+  build-push:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v4
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+      - name: Log in to GHCR
+        uses: docker/login-action@v4
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Compute image tags
+        id: image-tag
+        env:
+          INPUT_TAG: ${{ inputs.tag }}
+          MOVE_LATEST: ${{ inputs.move_latest }}
+        run: |
+          {
+            echo "tags<<EOF"
+            echo "ghcr.io/knaisoma/data-olympus:$INPUT_TAG"
+            if [ "$MOVE_LATEST" = "true" ]; then
+              echo "ghcr.io/knaisoma/data-olympus:latest"
+            fi
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: deploy/docker/Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: ${{ steps.image-tag.outputs.tags }}
+          labels: |
+            org.opencontainers.image.version=${{ inputs.tag }}
+            org.opencontainers.image.revision=${{ github.sha }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
--- a/.github/workflows/release-image.yml
+++ b/.github/workflows/release-image.yml
@@ -1,3 +1,4 @@
+# .github/workflows/release-image.yml
 name: Release Docker Image
 
 on:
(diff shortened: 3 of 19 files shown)


ajdelaguila (commented):
Required change before merge:

1. Add a `CHANGELOG.md` entry under the top `[Unreleased]` block for this PR. The PR changes observable release behavior: PR title enforcement, SemVer/release computation scripts, tag-on-version-bump automation, GitHub Release creation, and the image publishing path. `.rules/changelog-per-release.md` says every functional PR must update `[Unreleased]`, and release/CI behavior like this is user/operator-visible. The current CI changelog guard passed only because `scripts/check_changelog.py` does not treat `.github/workflows/` or release helper scripts as functional paths, but the project rule is broader than that implementation detail.

Feedback, not blocking:

- Consider adding a workflow-level `concurrency` group to `.github/workflows/tag-release.yml` with `cancel-in-progress: false`. Two close main pushes or a manual rerun can race on `git push origin "$TAG"`; one run will fail even though the workflow is intended to be idempotent.
- Decide whether publishing the GitHub Release before the image build succeeds is intentional. Today the release is visible before `build-image` completes, so a failed image build leaves a public release without its image until rerun.
- Please confirm the `uv.lock` marker churn is intentional and produced by the expected `uv` version. It appears unrelated to release automation and changes optional dependency resolution metadata.

Local verification I ran:
- `uv run pytest tests/test_compute_release.py tests/test_should_tag.py tests/test_lint_pr_title.py -q`
- `bats tests/bats/compute-version.bats`
- `uv run ruff check scripts/compute_release.py scripts/should_tag.py scripts/lint_pr_title.py tests/test_compute_release.py tests/test_should_tag.py tests/test_lint_pr_title.py`
- `git diff --check origin/main...HEAD`

ajdelaguila:
Thanks, all addressed in `07dadb9`.

**Required: CHANGELOG entry** — done. Added an `### Added` bullet under `[Unreleased]` describing the release pipeline (PR-title enforcement, SemVer/release scripts, tag-on-bump automation, Release creation, image path). You are right that the rule is broader than `check_changelog.py`'s functional-path set; the guard passing was an implementation detail, not license to skip the entry. The `changelog-guard` job now also has a real entry to point at.

**Feedback 1 (concurrency)** — done. Added a workflow-level `concurrency` group `tag-release-${{ github.ref }}` with `cancel-in-progress: false`, so two close main pushes or a rerun serialize instead of racing on `git push origin "$TAG"`.

**Feedback 2 (Release before image)** — changed to build-first. `tag-release.yml` is now three jobs: `decide` (create + push the annotated tag), `build-image` (`needs: decide`), and `release` (`needs: [decide, build-image]`) which creates the GitHub Release only after the image is published. A failed image build now leaves the tag but no public Release; a rerun is safe because tag creation and the Release step are both idempotent (`git rev-parse --verify` guard and `gh release view` skip).

**Feedback 3 (uv.lock churn)** — confirmed incidental and reverted. The churn was resolver marker-normalization on transitive CUDA/nvidia deps (from the `bench` extra), produced when a `uv run` refreshed the lock under uv 0.11.25. It was not required: I restored `origin/main`'s `uv.lock` and re-applied only the `data-olympus` `0.1.0 -> 0.1.1` version line, and `uv lock --check` passes. The `uv.lock` diff is now a single line.

Local verification: `uv run pytest` (release-script suites + full suite) green, `bats tests/bats/compute-version.bats` 5/5, `ruff check .` clean, both release workflows parse, `uv lock --check` consistent.

ajdelaguila (commented):
Remaining required change before merge:

The updated `tag-release.yml` still has an unrecoverable partial-release path. `decide` creates and pushes the tag, but `build-image` and `release` are both gated on `needs.decide.outputs.tag != ''`. On any full workflow rerun after the tag exists, `scripts/should_tag.py` sees the fetched `vX.Y.Z` tag and returns empty, because `tag_to_create()` returns `None` for any existing tag. That means both downstream jobs are skipped, so a release where the tag was pushed but the image build or GitHub Release step failed cannot be recovered by “Re-run all jobs” or by a later push to `main`. The idempotency guards in the tag and release steps never get a chance to run.

This directly contradicts the new comment that “a rerun is safe.” It is only safe for the narrow “Re-run failed jobs” path where GitHub can reuse the original successful `decide` output; a full rerun re-executes `decide` and loses the tag output.

Please decouple “desired version tag” from “new tag needed.” For example, have the decision step always emit the `pyproject.toml` tag when the tag is missing, or when the tag already points at the current commit but the image/release artifacts still need reconciliation. If the tag exists but points at a different commit, then no-op. That makes the downstream image/release idempotency live instead of unreachable.

Small doc nit while touching this: the new changelog entry says `tag-release.yml` “cuts the tag and Release ... then builds the image,” but the updated workflow intentionally does tag -> image -> GitHub Release.

Local verification after the update:
- `uv run ruff check .`
- `uv run mypy src`
- `uv run pytest -q` (1 skipped bench dependency test)
- `bats -r tests`
- `uv run data-olympus lint example-bundle`
- `git diff --check origin/main...HEAD`

I also noticed the PR check list currently shows only CodeQL checks on GitHub, so I relied on the local verification above for the repo CI surface.

ajdelaguila:
Good catch, you are right. Fixed in `ef21e6a`.

**The bug:** the idempotency lived in the tag/release steps, but a full rerun never reached them. `decide` re-ran `should_tag.py`, which returned empty for an existing tag, so `build-image` and `release` (both gated on `needs.decide.outputs.tag != ''`) were skipped. Only GitHub's "Re-run failed jobs" recovered, because it reuses the cached `decide` output. "Re-run all jobs" or a later push lost the tag output.

**The fix (in `scripts/should_tag.py`):** decoupled "desired tag" from "new tag needed", exactly as you suggested. New `release_action(version, existing, tag_commit, head_commit)`:

- tag missing -> emit `vX.Y.Z` (create, build, release)
- tag exists AND points at the current commit -> emit `vX.Y.Z` (reconcile: the downstream tag/image/release steps are all idempotent and finish whatever the failed run left undone)
- tag exists on a different commit, or its commit cannot be resolved -> no-op

`main()` now resolves the tag's commit (`git rev-list -n 1 <tag>`) and HEAD, and feeds them in. Because `tag-release.yml`'s `decide` job checks out with `fetch-depth: 0` (tags fetched), a full rerun on the release commit sees the tag pointing at HEAD and re-emits it, so `build-image` and `release` run and reconcile. The tag-creation step's `git rev-parse --verify` guard then skips re-tagging, and `git push origin "$TAG"` is a no-op. Once `main` advances past the release commit, the tag points at an older commit and this correctly no-ops.

Five new pure unit tests cover missing / at-HEAD / elsewhere / unresolved-commit. The workflow YAML itself did not need to change; the decision logic was the whole fix.

**Doc nit:** fixed. Corrected the ordering to "cuts the tag, builds the image, then publishes the GitHub Release" in both `CHANGELOG.md` and `.rules/versioning.md` (the same stale wording was in the rule file, so I updated it too).

Local verification: `uv run pytest tests/test_should_tag.py -q` 11/11, full `uv run pytest -q` green (1 bench skip), `uv run ruff check .` clean, changelog guard ok.

ajdelaguila (commented):
Rechecked the new `ef21e6a` changes. The remaining release-recovery blocker from my last review is addressed.

The important path now works: after a first run creates and pushes `vX.Y.Z` at the release commit but fails during image build or GitHub Release creation, a full workflow rerun checks out the same `github.sha`; `release_action()` sees that the existing tag points at `HEAD`, re-emits the tag, and the idempotent downstream `build-image` and `release` jobs run instead of being skipped. Steady state is preserved because later non-release pushes have `HEAD` after the release tag, so the decision emits nothing.

I do not have any remaining required changes from this pass. One non-blocking note: a full rerun of an already-complete release will rebuild and repush the same image/latest tag before `gh release view` short-circuits the Release step. That is safe but a little wasteful; acceptable as-is.

Local verification run after the update:
- `uv run pytest tests/test_should_tag.py -q`
- `uv run ruff check .`
- `uv run mypy src`
- `uv run pytest -q` (1 skipped bench dependency test)
- `bats -r tests` (passes; same expected Bats warning for command-not-found path test)
- `uv run data-olympus lint example-bundle`
- `git diff --check origin/main...HEAD`

GitHub PR checks still show only CodeQL jobs, so the repo CI surface above was verified locally.

PR #60, opened 2026-07-02
feat(benchmarks): content-free real-corpus retrieval eval harness

ajdelaguila:
## Summary

Ships a **content-free** eval harness so anyone can measure, on their own KB,
what the optional local-embedding hybrid (#42) adds over the lexical stack
(FTS + synonym + co-occurrence expansion). Docs/benchmarks only — no runtime
behavior change.

- `benchmarks/real_corpus_eval.py` — point it at a corpus dir + a labeled
  queries JSON (`[{text, gold_ids}]`); it builds a lexical-stack index and a
  hybrid (+embeddings) index over a temp DB and reports recall@k, MRR, and
  recovered/regressed counts. It prints **only aggregate metrics, never document
  text**, so it is safe to run over a private KB.
- `benchmarks/real_corpus_eval.md` — usage, a worked private-corpus example, and
  an honest reading.

## Worked example (our internal KB, aggregate only)

243 indexed docs, 215 natural-language **paraphrase** queries (one per doc,
worded to avoid the doc's title terms), default config:

| Metric | Lexical stack | Hybrid (+embeddings) | Delta |
|---|---|---|---|
| recall@5 | 0.898 | 0.916 | +0.019 |
| recall@10 | 0.953 | 0.967 | +0.014 |
| MRR | 0.756 | 0.770 | +0.013 |

Hybrid recovered 4/215 (three token-disjoint) and regressed 0/215.

## Honest reading

The lexical stack is already strong (~0.90 recall@5) even on deliberately
paraphrased queries, so headroom is small. The embedding hybrid is a real but
small, strictly non-harmful add here — which is why it ships **opt-in**
(`KB_EMBEDDINGS_MODE`), not default-on. Enable it when your corpus has genuine
semantic gaps, and measure with this harness first.

No KB documents are published — only the generic harness code and aggregate
numbers.

### DIFF
--- a/benchmarks/real_corpus_eval.py
+++ b/benchmarks/real_corpus_eval.py
@@ -22,18 +22,10 @@ from __future__ import annotations
 import argparse
 import json
 import statistics
+import sys
 from pathlib import Path
 
-
-def recall_at_k(ranked: list[str], gold: set[str], k: int) -> float:
-    return 1.0 if gold and any(r in gold for r in ranked[:k]) else 0.0
-
-
-def mrr(ranked: list[str], gold: set[str]) -> float:
-    for i, r in enumerate(ranked, 1):
-        if r in gold:
-            return 1.0 / i
-    return 0.0
+from benchmarks.metrics import mrr, recall_at_k
 
 
 def _build_expanders(index):  # noqa: ANN001 - Index, avoid import at module load
@@ -68,6 +60,8 @@ def main() -> None:
     raw = json.loads(Path(args.queries).read_text(encoding="utf-8"))
     queries = [q for q in raw if q.get("gold_ids")]
     print(f"corpus={corpus} queries={len(raw)} (labeled={len(queries)}) k={args.k}")
+    if not queries:
+        sys.exit("no labeled queries found: every entry has empty/missing gold_ids")
 
     with tempfile.TemporaryDirectory() as tmp:
         tmp_path = Path(tmp)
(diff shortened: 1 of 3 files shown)


benchmarks/real_corpus_eval.py
```diff
@@ -0,0 +1,116 @@
...
+"""
+from __future__ import annotations
+
+import argparse
+import json
+import statistics
+from pathlib import Path
+
+
+def recall_at_k(ranked: list[str], gold: set[str], k: int) -> float:
```
ajdelaguila: Nice to have: this local `recall_at_k` uses binary any-hit semantics, while the repo's canonical `benchmarks.metrics.recall_at_k` is fractional for multiple `gold_ids`. The worked example uses one gold id per query, so the current numbers are unaffected, but importing the canonical helper or renaming this to `hit_rate_at_k` would avoid future metric confusion.

benchmarks/real_corpus_eval.py
```diff
@@ -0,0 +1,116 @@
...
+            hr = [h.id for h in hyb.search(q["text"], limit=args.k)]
+            lrec, hrec = recall_at_k(lr, gold, args.k), recall_at_k(hr, gold, args.k)
+            lex_recall.append(lrec)
+            hyb_recall.append(hrec)
+            lex_mrr.append(mrr(lr, gold))
+            hyb_mrr.append(mrr(hr, gold))
+            recovered += int(hrec > lrec)
+            regressed += int(hrec < lrec)
+
+    lr_m, hr_m = statistics.mean(lex_recall), statistics.mean(hyb_recall)
```
ajdelaguila, line 101: Nice to have: if every query has empty or missing `gold_ids`, `queries` is empty and `statistics.mean([])` raises an opaque `StatisticsError`. A small early guard with a clear message like `no labeled queries found` would make this easier to diagnose.

benchmarks/real_corpus_eval.py
```diff
@@ -0,0 +1,116 @@
...
+    lm_m, hm_m = statistics.mean(lex_mrr), statistics.mean(hyb_mrr)
+    print("\n== lexical stack vs hybrid (+embeddings), default config ==")
```
ajdelaguila, line 103: Nice to have: because the searches above use `limit=args.k`, the reported MRR is really MRR@k. Clarifying the label in the output/docs, or computing MRR from a wider untruncated retrieval, would make the metric easier to interpret across different `--k` runs.

ajdelaguila (commented):
Non-blocking review notes only. These are nice-to-have cleanups for metric clarity and edge-case handling; no requested changes.

PR #149, opened 2026-07-12
feat: 0.5.0 release batch of 6 features (KNA-66)

### LINKED ISSUE #69: Ops: Prometheus metrics endpoint
## Summary
A minimal Prometheus exposition endpoint (`/metrics`) on the REST server: pending-queue depth, push-queue depth and frozen count, push failure counter, staleness seconds, live MCP session count, per-tool call counters, index build duration and last-build timestamp.

## Motivation
Current observability is a health JSON plus logs. The health surface also under-reports today (write-side counters are not wired), so operators cannot see stuck pushes, queue growth, or worktree accumulation at all. Alerting on a governance-critical single-writer service currently means grepping logs.

## Sketch
- Gauge/counter registry updated in the existing loops (refresh, push retry, pending GC) and tool wrappers; no new deps beyond `prometheus-client` (optional extra).
- Ship example alert rules in the ops runbook (staleness > N, frozen pushes > 0, pending > N for M hours).

From the 2026-07-03 project audit (docs/plans/2026-07-03-project-audit-action-plan.md, sections 3.6 and 6).

### LINKED ISSUE #137: Consult demoted_writes CTA counts resolved demotions as awaiting review
## Summary

The `pending_actions` item of kind `demoted_writes` on `kb_consult` says "N write(s) in this session are awaiting operator review", but N appears to be the session-lifetime demotion count from the audit log, not the number of still-unresolved pending entries.

## Reproduction (observed on kn-dev, v0.4.0, 2026-07-08)

- A high-confidence `kb_propose_edit` to an in-force doc was demoted (governed-lane, `demotion_reason: status_promotion`).
- The pending entry was then resolved with `decision: reject`.
- `kb_health` shows `pending_count: 0`; `kb_list_pending` is empty.
- A subsequent `kb_consult` for the same `source_session` still returns: `{"kind": "demoted_writes", "message": "1 write(s) in this session are awaiting operator review...", "count": 1}`.

## Expected

Either the count reconciles against live pending entries for the session (0 here, so the item is omitted), or the wording changes to reflect what is actually counted (e.g. "were demoted this session; M still awaiting review"). The session recap surface (`kb_session_recap`) counting lifetime demotions is fine and intended; the CTA's "awaiting" claim is what mismatches.

Found during the v0.4.0 post-deploy verification pass.

### LINKED ISSUE #139: Serving: first-class reverse-proxy host allowlist for fastmcp DNS-rebinding protection
## Summary

fastmcp 3.4.3 (shipped in data-olympus v0.4.1) enables streamable-HTTP DNS-rebinding protection: it validates the Host/Origin headers and returns 421 Misdirected Request for hostnames outside its allowlist. Any deployment behind a reverse proxy or ingress (the normal production shape) breaks entirely unless `FASTMCP_HTTP_ALLOWED_HOSTS` is set, while direct pod probes keep passing, so readiness stays green and the failure is invisible to the orchestrator.

Observed on kn-dev 2026-07-09: every ingress request 421'd for ~7 hours; enforcement hooks failed open; kubelet /readyz probes returned 200 the whole time.

## Proposed

- A first-class config knob (e.g. `KB_PUBLIC_HOSTNAMES`) mapped to fastmcp's `allowed_hosts` when building the HTTP app, so operators configure data-olympus, not its dependency, and misconfiguration is loggable at startup.
- A startup WARN when the server binds non-localhost with host protection on and no extra allowed hosts configured (the exact silent-breakage shape).
- Consider surfacing 421-rejection counts in kb_health so a host-mismatch is observable from the inside.
- Regression test: app boots with a configured public hostname; a request carrying that Host header succeeds; a foreign Host still 421s.

## Workaround (documented)

Set `FASTMCP_HTTP_ALLOWED_HOSTS: '["<public-hostname>"]'` in the deployment env (JSON list, parsed by pydantic-settings). Applied to the kn-dev configmap.

## Postmortem notes

- The dependabot bump's CI was green: tests exercise the app directly and never send a proxied Host header, exactly the blind spot the regression test above closes.
- The kn-dev statefulset carries keel.sh/policy: minor (2-minute poll), so the broken patch auto-deploys to any keel-managed environment; a plain image rollback is reverted by Keel within minutes. Rollbacks there require pausing/adjusting the Keel policy first.

### LINKED ISSUE #146: Notify operator/model when a newer data-olympus version is published
## Problem

Deployments that do not auto-update (i.e. anyone not running Keel or similar) can silently run a stale `data-olympus` version, unaware that a newer release with fixes or features has been published. There is no in-band signal that "you are behind."

## Feature

When the running server (and/or CLI) detects that a newer version has been published, surface a non-intrusive "a new version X.Y.Z is available" notice to the human operator and/or the model. Candidate surfaces (pick during design): a field on `GET /api/v1/health`, an annotation on MCP tool responses, a one-time startup log line, and/or `data-olympus setup --check` output.

## Prior art

`src/data_olympus/setup_wizard.py::latest_version()` already looks up the latest published version from PyPI (and GitHub), tolerant of being offline, and is used by `data-olympus setup --check`. This feature extends that lookup to a runtime/operational surface.

## Considerations

- Opt-out / offline-safe (air-gapped deployments must not error or block).
- Cache + rate-limit the check (do not hit PyPI on every request; e.g. once per N hours).
- Privacy: this is a read of the public latest version only, not telemetry.
- Not needed for Keel-auto-update deployments; valuable for manual ones.

Part of the 0.5.0 batch.

### LINKED ISSUE #147: Auto-add a default status to corpus docs missing it (pre-0.4.0 migration)
## Problem

Since 0.4.0, a corpus document missing `status` is a `kb lint` error: it is still served but **never treated as in-force** (search/enforcement stop treating it as authoritative), tracked in the maintenance ledger until an operator adds `status` (see `tests/test_status_mandatory_migration.py`, `maintenance.py`, issue #114). A corpus upgraded from **pre-0.4.0** therefore has its status-less documents silently drop out of the in-force set on upgrade.

## Decision (operator, 2026-07-11): option (a) — auto-add a safe default

Auto-add a safe default `status` to documents missing it, so legacy corpuses migrate seamlessly instead of silently losing in-force status.

## Proposed approach

- On index build / a dedicated migration pass, for any doc missing `status`, write a default `status` via the normal serialized write/commit machinery (idempotent, audited) — not an agent write, so it must bypass the governed-lane pending route.
- **Key design point: the default value.** `active` preserves pre-0.4.0 in-force behavior (backward-compat, the goal here) but auto-promotes legacy docs to authoritative; a more conservative default keeps them served-but-flagged. Finalize the default and whether it is config-gated (e.g. `KB_STATUS_AUTOFILL=on`).
- This intentionally reverses the original "never guess status" stance (#114); document the rationale (seamless upgrade > conservative flagging) in SPEC.md / `.rules`.

## Acceptance

- A corpus with status-less docs, after this runs, has every doc carrying `status`, and the previously-in-force legacy docs remain in-force.
- Idempotent; audited; covered by a migration test extending `test_status_mandatory_migration.py`.

Part of the 0.5.0 batch.

### LINKED ISSUE #148: Block publishing a version that already exists on an external registry
## Rule

Once a version has been published to an external registry (PyPI, ghcr, GitHub Releases), it is **immutable**: we must never publish a modified artifact under the same version number. The release process must make this impossible, not just discouraged.

## Feature

A strict pre-publish guard in the release pipeline that queries the external registries for the target version and **hard-fails** the release if that version already exists. Complements `scripts/should_tag.py`'s idempotency (which only re-emits a tag when it points at the current commit).

## Proposed approach

- A small script `scripts/check_version_free.py --version X.Y.Z` (mirroring `scripts/security_alerts.py`) that checks PyPI (`/pypi/data-olympus/<v>/json`), ghcr (`:v<X.Y.Z>` tag), and the GitHub release/tag; exits non-zero if the version already exists anywhere.
- Wire it as a gate in `tag-release.yml` before the publish jobs, and as a cutter readiness step in `.rules/release-routine.md`.
- Distinguish "already published" (hard block) from the legitimate idempotent partial-failure re-run (same commit/digest -> allowed to finish).

Part of the 0.5.0 batch.

ajdelaguila:
## Release 0.5.0 integration batch (epic KNA-66)

Implements the 6 operator-approved features for data-olympus 0.5.0. Scope, per-issue specs, security clearance, and the dual-architect review are recorded in Paperclip issue KNA-65 (release-scope-brief). Bump 0.4.2 -> 0.5.0 (minor).

Security clearance PASSED at planning time: 0 open Dependabot / CodeQL / secret-scanning alerts.

### Features (one commit each)

- Closes #148 - Version-immutability release guard (scripts/check_version_free.py); wired into tag-release.yml before the tag push and into PR CI; fails closed on registry outage with KB_BYPASS_VERSION_CHECK override.
- Closes #137 - Fix demoted_writes CTA in kb_consult to reconcile against the live pending queue (omits the item once a session's demotions are resolved).
- Closes #139 - First-class reverse-proxy Host allowlist (KB_PUBLIC_HOSTNAMES) mapped onto fastmcp allowed_hosts, with a startup WARN and a regression test for the 421 blind spot.
- Closes #69 - Optional Prometheus /metrics endpoint; metrics registry is a no-op stub when prometheus-client is absent (background loops never crash).
- Closes #146 - Notify when a newer version is published; runs as a periodic background task (never on the async health request path), cached on ServerState, gated by KB_DISABLE_VERSION_CHECK.
- Closes #147 - Virtual status autofill for legacy corpora (KB_STATUS_AUTOFILL, default on, in-memory only so Index.build stays read-only) plus an explicit data-olympus migrate status --apply persistence lane.

### Verification (integrated tree)

- uv run pytest: 1616 passed, 2 skipped (optional deps).
- uv run ruff check .: clean. uv run mypy src: clean (72 files).
- check_doc_consistency.py, check_benchmark_docs.py, data-olympus lint example-bundle, and bats -r tests: all pass.

Delivery model: WF-004 section 2.2 shared integration branch feature/KNA-66, one commit per feature, this single integration MR merge-commits to main.


ajdelaguila:
## Code review: KNA-68 slice (issue #146 — notify operator/model when a newer version is published)

Scope: `src/data_olympus/{config,health,models,rest_api,server,tools_read,version_check}.py`, `tests/test_version_check.py`, `CHANGELOG.md`, `SPEC.md`, `docs/{serving,operations}.md`, at `960115f` (feature/KNA-66). This is a scoped review of the KNA-68 slice of this integration PR, not the full 6-feature batch.

**Design correctness**
- `version_check.py`: the blocking `setup_wizard.latest_version()` call is correctly kept off the event loop via `asyncio.to_thread`; the `/api/v1/health` route reads only the cached `ServerState.latest_version`/`update_available` fields, so a slow/unreachable PyPI can't stall readiness. Confirmed by reading `rest_api.py`/`server.py` wiring, not just the docstring claim.
- `KB_DISABLE_VERSION_CHECK` gates task creation in `_maybe_start_version_check_task`, so an air-gapped deployment spawns zero background work — verified this is a hard `return`, not just a no-op inside the loop.
- One-time warn-on-detect (`warned` flag) avoids re-logging every interval once a newer version is found; `asyncio.CancelledError` is re-raised (clean shutdown), other exceptions are swallowed with a log line so a transient failure doesn't kill the loop.
- The `run_once_immediately=False` deferral path in `version_check_loop` (skip the immediate check, wait one interval) traced correctly against the `first`/`state.latest_version is None` guard.

**Verification (independently reproduced, not just taken from the ticket)**
- `uv run pytest tests/test_version_check.py -q` → 14 passed, in `.worktrees/KNA-66-integrate` at `960115f`.
- Diff reviewed in full against `main...feature/KNA-66` for the listed files; config/health/models/rest_api/server wiring matches the stated intent (cache-only health read, config knobs default to current behavior, `disable_version_check` default `False` / `version_check_interval_sec` default `86400`).

**Findings:** none blocking. No correctness, security, or reuse issues found in this slice.

**Release dependency (flagging per KNA-85 scope):** this PR (#149) bundles 6 features for the 0.5.0 release batch (KNA-66); merge readiness depends on the other slices' reviews and on PR CI, which had the `test` job still running as of this review. Not a blocker on this slice's review, but noting it since this PR itself isn't mergeable until CI is green and KNA-66 integration is otherwise ready.

PR #167, opened 2026-07-18
fix(release): build images from requested ref

ajdelaguila:
## Summary

Fix RC and release image provenance so the reusable image workflow builds the explicit requested ref instead of the caller workflow SHA.

## Verification

* `uv run --python 3.13 --extra dev pytest -q tests/test_publish_workflows.py`: 12 passed
* `actionlint`: passed
* Independent Claude high risk review: APPROVE at `33b05a208c60e89b17db825242b0041dc47f4f57`

The operator supplied release ref is consumed by `actions/checkout`, not interpolated into shell. The resolved commit is obtained with `git rev-parse HEAD` after checkout and reused immutably by image and prerelease jobs.

RC.1 exposed this defect and was not deployed. A new RC will be cut only after this correction lands on the trusted `main` workflow definition.

### DIFF
--- a/.github/workflows/rc-publish.yml
+++ b/.github/workflows/rc-publish.yml
@@ -26,11 +26,15 @@ jobs:
     outputs:
       rc_tag: ${{ steps.rc.outputs.rc_tag }}
       version: ${{ steps.rc.outputs.version }}
+      source_sha: ${{ steps.source.outputs.sha }}
     steps:
       - uses: actions/checkout@v7
         with:
           ref: ${{ inputs.ref }}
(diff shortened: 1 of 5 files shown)


ajdelaguila:
Independent high risk review: APPROVE at `33b05a208c60e89b17db825242b0041dc47f4f57`.

The reviewer confirmed the fix closes both the observed decision/build divergence and the broader moving ref race by resolving once and pinning every downstream consumer. The requested ref is passed only to `actions/checkout`; no user controlled ref is interpolated into shell. Exact SHA CI is fully green.

PR #238, opened 2026-08-16
fix(deploy): bootstrap /kb-main when only KB_REMOTE_URL is set

yuzi-co:
## The problem

`deploy/docker/entrypoint.sh` keys its first-boot clone off `KB_GIT_REMOTE_URL`:

```sh
if [ -n "${KB_GIT_REMOTE_URL:-}" ] && [ -d /kb-main ]; then
```

But `deploy/docker/compose.yaml` carries only `KB_REMOTE_URL`, which is also the variable documented in `docs/adoption.md` and `docs/quickstart.md`, and the one the server reads for its push target.

So on the compose path the block is a no-op, and a deployment that configured a remote gets a **read-write server on an unbootstrapped `/kb-main`** — even though the block's own comment says it handles the clone "for a plain `docker compose` run (no initContainer)".

## Why it is hard to diagnose

Nothing says "not cloned". The MCP write tools return

```json
{"status": "write_pipeline_disabled"}
```

as a *successful* call rather than an error, and the visible symptom is

```
git_pull_loop iteration failed: Command '['git', '-C', '/kb-main',
'rev-parse', 'HEAD']' returned non-zero exit status 128
```

which names neither the missing clone nor the variable responsible.

## The fix

The two names look like a mistake but are a deliberate split: `KB_GIT_REMOTE_URL` is the clone source read by the entrypoint and the k8s initContainer, `KB_REMOTE_URL` is the push target read by the server, and `deploy/k8s/statefulset.yaml` wires **both** from a single secret key. So this renames nothing. It falls back:

```sh
clone_url="${KB_GIT_REMOTE_URL:-${KB_REMOTE_URL:-}}"
```

`KB_GIT_REMOTE_URL` still wins when set, leaving the k8s path byte-identical, while the compose path bootstraps as documented.

The pre-clone log line no longer prints the URL (review follow-up in `3ce063a`). `KB_REMOTE_URL` is documented as an SSH or HTTPS push target and the compose path mounts no SSH key by default, so the realistic way to get a writable remote there is an HTTPS URL with an embedded token. Routing that variable into the clone would have put the token in the container log, before the clone runs and regardless of whether it succeeds, where no application-level redaction can reach it. The message now reads:

```sh
echo "[entrypoint] /kb-main is empty; cloning configured remote"
```

That closes the unconditional disclosure, not every path to one. `git` inherits the same stdout and stderr, and a token embedded in a clone URL is written to `/kb-main/.git/config` once the clone succeeds. Both follow from putting a credential in the URL at all, and provisioning it separately is the way out; neither is in scope here.

What stays unchanged on the k8s path is remote selection: `deploy/k8s/statefulset.yaml` and `deploy/k8s/read-replica/deployment.yaml` bind both variables to the same secret key, so the `:-` fallthrough cannot pick a different remote there. The line that path would log if it ever did bootstrap does change.

## Verification

Ran the isolated bootstrap block against a local bare repo in all three cases:

| environment | result |
|---|---|
| only `KB_REMOTE_URL` set | clones — this is the case that was broken |
| both set | uses `KB_GIT_REMOTE_URL`, k8s behaviour unchanged |
| neither set | no clone, exits 0 |

Those three cases plus the log-redaction rule are now regression tests in `tests/test_deploy_manifests.py`. They run the real block rather than grepping it: the block is lifted out of `entrypoint.sh` between its marker comment and `exec "$@"`, with only the `/kb-main` mount point redirected into a tmp dir, and executed with a stub `git` on `PATH` that records its argv and prints nothing. So the assertions cover which URL git was actually invoked with, and the captured output is entirely the entrypoint's own.

Every test was checked against a mutated `entrypoint.sh`, since a test that has never failed proves nothing:

| mutation | result |
|---|---|
| `cloning ${clone_url}` | caught (2 tests) |
| `cloning ${KB_GIT_REMOTE_URL}` | caught (1) |
| `cloning ${KB_REMOTE_URL}` | caught (1) |
| fallback removed | caught (2) |
| precedence inverted | caught (1) |

`sh -n` clean, `uv run ruff check .` clean, `uv run mypy src` clean, `uv run pytest` exit 0 (4 environment skips: `sentence_transformers`, `bun` x2, `tiktoken`) on a fresh clone of this branch at `2469aef`, installed with `uv pip install -e '.[dev]'` as CI does.

Rebased onto `a1913ad`. The only conflict was `CHANGELOG.md`; both entries now sit under the current empty `[Unreleased]` block, above `## [0.7.2] - 2026-09-07`.

Found while deploying v0.7.1 via `docker compose` against a self-hosted remote.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,19 @@ This project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+* **Bootstrap `/kb-main` on the docker compose path.** The first-boot clone in
+  `deploy/docker/entrypoint.sh` keyed off `KB_GIT_REMOTE_URL`, but
+  `deploy/docker/compose.yaml` carries only `KB_REMOTE_URL` — which is also the
+  variable `docs/adoption.md` and `docs/quickstart.md` document. A compose
+  deployment that configured a remote therefore started a read-write server on
+  an empty `/kb-main`, and the first symptom was
+  `git_pull_loop iteration failed: … rev-parse HEAD … exit status 128`, which
+  does not point at the cause. The clone now falls back to `KB_REMOTE_URL`;
+  `KB_GIT_REMOTE_URL` still takes precedence, so the k8s path where the
+  initContainer has already cloned is unchanged.
+
 ## [0.7.2] - 2026-09-07
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,19 @@ This project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   `KB_GIT_REMOTE_URL` still takes precedence, so the k8s path where the
   initContainer has already cloned is unchanged.
 
+### Security
+
+* **The first-boot clone no longer logs the remote URL.** `KB_REMOTE_URL` is
+  documented as an SSH or HTTPS push target and the compose path mounts no SSH
+  key by default, so a writable remote there is most likely an HTTPS URL with an
+  embedded token. The entrypoint printed that URL verbatim before cloning, into
+  a container log no application-level redaction can reach. It now reports only
+  that it is cloning the configured remote. That removes the unconditional
+  disclosure rather than every path to one: git inherits the same stdout and
+  stderr, and a token embedded in a clone URL is still written to
+  `/kb-main/.git/config` once the clone succeeds. Provisioning the credential
+  outside the URL is the way out of both.
+
 ## [0.7.2] - 2026-09-07
 
 ### Changed
@@ -48,6 +61,7 @@ This project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   and fresh zero-alert clearance after the merge before anything is published.
   Previously the rule read as an unqualified zero-alert gate, which no branch
   fixing a default-branch alert could ever satisfy before merging.
+
 
 ### Fixed
 
--- a/deploy/docker/compose.yaml
+++ b/deploy/docker/compose.yaml
@@ -16,6 +16,10 @@ services:
       KB_HTTP_PORT: "8080"
       KB_SYNC_INTERVAL_SEC: "60"
       KB_STALENESS_DEGRADED_SEC: "600"
+      # Set this to a git remote to enable the write pipeline; left empty the
+      # server runs read-only. The entrypoint also uses it to clone /kb-main on
+      # first boot when KB_GIT_REMOTE_URL is not set separately, so on this
+      # path one variable is enough.
       KB_REMOTE_URL: ""
     volumes:
       - kb-main:/kb-main
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -67,14 +67,28 @@ fi
 # --- Bootstrap /kb-main on first boot (docker compose path only) -------------
 # In k8s the initContainer already cloned /kb-main. For a bare docker run there is
 # no initContainer, so do it here. Idempotent: skipped when /kb-main is non-empty.
-# Safe: only clones when KB_GIT_REMOTE_URL is set. On clone failure the volume
-# stays empty and the MCP reports degraded so the operator notices.
-if [ -n "${KB_GIT_REMOTE_URL:-}" ] && [ -d /kb-main ]; then
+# On clone failure the volume stays empty and the MCP reports degraded so the
+# operator notices.
+#
+# Two variables name a remote: KB_GIT_REMOTE_URL is the clone source read here
+# and by the k8s initContainer, KB_REMOTE_URL is the push target read by the
+# server. The k8s manifests set both from one secret key, but compose.yaml
+# carries only KB_REMOTE_URL -- which is also the one docs/adoption.md and
+# docs/quickstart.md document. A compose deployment that set the documented
+# variable therefore got a read-write server on an unbootstrapped /kb-main, and
+# the first symptom was `git_pull_loop iteration failed: ... rev-parse HEAD ...
+# exit status 128`, which does not point at the cause.
+#
+# Falling back keeps k8s behaviour identical -- KB_GIT_REMOTE_URL still wins
+# when set -- while making the compose path bootstrap as this block's own
+# comment already claims it does.
+clone_url="${KB_GIT_REMOTE_URL:-${KB_REMOTE_URL:-}}"
+if [ -n "$clone_url" ] && [ -d /kb-main ]; then
     found_real_files=$(find /kb-main -mindepth 1 -maxdepth 1 \
         ! -name 'lost+found' -print -quit 2>/dev/null || true)
     if [ -z "$found_real_files" ]; then
-        echo "[entrypoint] /kb-main is empty; cloning ${KB_GIT_REMOTE_URL}"
-        if git clone "$KB_GIT_REMOTE_URL" /kb-main; then
+        echo "[entrypoint] /kb-main is empty; cloning ${clone_url}"
+        if git clone "$clone_url" /kb-main; then
             echo "[entrypoint] bootstrap complete"
         else
             echo "[entrypoint] WARNING: clone failed; pod will start degraded"
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -87,7 +87,13 @@ if [ -n "$clone_url" ] && [ -d /kb-main ]; then
     found_real_files=$(find /kb-main -mindepth 1 -maxdepth 1 \
         ! -name 'lost+found' -print -quit 2>/dev/null || true)
     if [ -z "$found_real_files" ]; then
-        echo "[entrypoint] /kb-main is empty; cloning ${clone_url}"
+        # The URL is deliberately absent from this message. KB_REMOTE_URL is
+        # documented as an "SSH or HTTPS" push target and this path mounts no SSH
+        # key by default, so the realistic way to reach a writable remote here is
+        # an HTTPS URL with an embedded token. Echoing it would put that token in
+        # the container log before the clone even runs, where nothing downstream
+        # can redact it. The operator already has the value in the environment.
+        echo "[entrypoint] /kb-main is empty; cloning configured remote"
         if git clone "$clone_url" /kb-main; then
             echo "[entrypoint] bootstrap complete"
         else
--- a/tests/test_deploy_manifests.py
+++ b/tests/test_deploy_manifests.py
@@ -165,3 +165,24 @@ def test_entrypoint_has_no_gosu_or_chown() -> None:
 def test_compose_binds_loopback() -> None:
     text = (DOCKER_DIR / "compose.yaml").read_text()
     assert "127.0.0.1:8080:8080" in text
+
+
+def test_entrypoint_bootstrap_falls_back_to_kb_remote_url() -> None:
+    """The compose path sets only KB_REMOTE_URL -- it is the documented one and
+    the only remote variable compose.yaml carries -- so the first-boot clone
+    must fall back to it. Without the fallback a compose deployment that
+    configured a remote started a read-write server on an unbootstrapped
+    /kb-main. KB_GIT_REMOTE_URL must still take precedence so the k8s path,
+    where the initContainer has already cloned, is unchanged.
+    """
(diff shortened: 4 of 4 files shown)


ajdelaguila:
Thanks for tracking this one down, and for explaining why it is hard to diagnose. I confirmed the premise on current `main`: `deploy/docker/compose.yaml` carries only `KB_REMOTE_URL`, while the first-boot clone in `deploy/docker/entrypoint.sh` keys off `KB_GIT_REMOTE_URL`, so on the compose path that block is dead code and a deployment that configured a remote does start a read-write server on an unbootstrapped `/kb-main`.

Your read of the two-variable split is also right: they are deliberately distinct (clone source vs push target) and `deploy/k8s/statefulset.yaml` wires both from one secret key, so the fallback is the correct shape of fix and leaves the k8s path byte-identical. Reviewed at `a84739b`.

One blocker before this can go in.

## Blocker: the retained log line can now print a credential

```sh
clone_url="${KB_GIT_REMOTE_URL:-${KB_REMOTE_URL:-}}"
...
echo "[entrypoint] /kb-main is empty; cloning ${clone_url}"
```

`KB_REMOTE_URL` is documented as the push target, "SSH or HTTPS". The compose file mounts neither an SSH key nor a credential helper by default, so the realistic way to get a writable remote on that path is an HTTPS URL with an embedded token. This line prints it verbatim into container logs, before the clone and regardless of whether the clone then succeeds. Nothing downstream redacts it, and application-level redaction cannot reach output the shell already emitted.

To be fair to the existing code: that echo was already unsafe for a credential-bearing `KB_GIT_REMOTE_URL`. But on the k8s path that variable is used with a mounted SSH key, whereas this change routes the documented push-target variable into it on the one path with no key mount, which is where an inline token is most likely. So the exposure becomes materially more reachable.

Simplest fix is to drop the URL from the message entirely, since it adds nothing the operator cannot get from the environment:

```sh
echo "[entrypoint] /kb-main is empty; cloning configured remote"
```

## The two new tests only check that a string appears in a file

`test_entrypoint_bootstrap_falls_back_to_kb_remote_url` asserts the literal expansion is present in `entrypoint.sh`, and `test_compose_documents_that_kb_remote_url_also_bootstraps` asserts `KB_GIT_REMOTE_URL` appears somewhere in `compose.yaml`. The first passes even if the expansion were sitting in a comment or an unused assignment, and neither exercises clone behaviour or the precedence rule they describe.

They match the existing style in that file, so this is not a blocker. But the behaviour you already verified by hand (fallback, precedence, neither set) is the part worth locking down, and it is reachable in a test with a stub `git` on `PATH` asserting which URL it was invoked with, plus an assertion that the emitted log contains no credential.

## CI

Workflows here are at `action_required` because this is a fork PR, so nothing has run and the ruff/pytest results in the description are unverified on our side. Approving that run is our action; we will get checks going once the log line is addressed.

yuzi-co:
Both addressed in 9a721af.

## Blocker: the log line

Dropped the URL, as suggested:

```sh
echo "[entrypoint] /kb-main is empty; cloning configured remote"
```

with a comment above it recording why, so the next person does not helpfully add the URL back.

One thing I checked before claiming the leak is closed: git's own output does not put the credential back. Against a credential-bearing HTTPS URL, `git clone` prints

```
fatal: unable to access 'https://127.0.0.1:1/x.git/': Failed to connect ...
```

with the userinfo stripped, so the entrypoint's echo really was the only place the token appeared on that path.

I left `deploy/k8s/statefulset.yaml` alone. Its initContainer has the same shape of line (`echo "[init] /kb-main empty; cloning $KB_GIT_REMOTE_URL"`), but it is pre-existing, that path mounts an SSH key, and touching it would cost the "k8s stays byte-identical" property this PR leans on. Happy to send it as a separate PR if you want it closed too.

## Tests

Replaced `test_entrypoint_bootstrap_falls_back_to_kb_remote_url` with four tests that run the block instead of grepping it. The compose doc assertion stayed as a text check, since a comment existing in a YAML file is all it claims.

How they run it: the block is lifted out of `entrypoint.sh` verbatim, between its marker comment and `exec "$@"`, with only the `/kb-main` mount point rewritten to a tmp dir (the container mount cannot exist on a test host). A stub `git` earlier on `PATH` appends its argv to a file and prints nothing, so anything captured is the entrypoint's own output. Running the whole entrypoint was not an option: it writes fixed absolute paths such as `/tmp/known_hosts` and ends in `exec "$@"`.

What they assert:

- only `KB_REMOTE_URL` set: git is invoked with `clone` and that URL
- both set: the clone source is `KB_GIT_REMOTE_URL`, and the push target is not passed to git
- neither set: git is never invoked and nothing is logged
- credential-bearing `KB_REMOTE_URL`: the token and host are absent from the output, with an assertion that the clone still used the URL so the test cannot pass vacuously

Mutation-checked, since a test that has never failed proves nothing:

| mutation | result |
|---|---|
| restore `cloning ${clone_url}` | credential test fails |
| drop the fallback | fallback and credential tests fail |
| invert the precedence | precedence test fails |

They are skipped where `os.name != "posix"`, so they run on the CI runner and on any developer machine with a POSIX shell.

## CHANGELOG

Added a `### Security` entry for the log line, since it changes what an existing deployment prints at boot.

## CI

Re-ran the checks locally on 9a721af, from a fresh clone with `uv pip install -e '.[dev]'` as CI does: `sh -n` clean, `uv run ruff check .` clean, `uv run pytest` exit 0 with 4 environment skips (`sentence_transformers`, `bun` x2, `tiktoken`). Still unverified on your side until the fork run is approved.

ajdelaguila:
The blocker is resolved in `9a721af`, and the rewritten tests are a real improvement over the string match.

What I checked by reading the source, not by running it:

- The pre-clone message no longer interpolates a URL, and `test_bootstrap_log_does_not_leak_a_credential` asserts that the token and the host are both absent from the block's output while still asserting that the clone used that URL, so it cannot pass vacuously.
- Precedence is right and the manifests stay untouched. `deploy/k8s/statefulset.yaml` and `deploy/k8s/read-replica/deployment.yaml` both bind `KB_GIT_REMOTE_URL` and `KB_REMOTE_URL` to the same secret key, so the `:-` fallthrough cannot select a different remote there. Worth stating precisely, though: what is unchanged on that path is remote selection, not every behaviour, since a bootstrap attempt there would now log a different line.
- The harness matches production shell options. `entrypoint.sh` is `#!/bin/sh` with `set -e`, and the extracted block runs under exactly `set -e`.
- Your `git clone` observation reproduces here. Against a credential-bearing HTTPS URL, git 2.50.1 prints `fatal: unable to access 'https://127.0.0.1:1/x.git/': ...` with the userinfo stripped.

One thing I would like before this goes in, and it is not a blocker:

**The four tests would not catch the regression they exist to prevent.** Restoring `echo "[entrypoint] /kb-main is empty; cloning ${KB_GIT_REMOTE_URL}"` passes all four of them. The credential test sets only `KB_REMOTE_URL`, so `${KB_GIT_REMOTE_URL}` expands empty there and nothing leaks; the precedence test does set it, but with non-secret URLs and without looking at the output at all. Your mutation table covers `cloning ${clone_url}`, which is the post-fix spelling, so it does not reach this shape. A second credential case that puts the token in `KB_GIT_REMOTE_URL` closes the gap.

Two smaller notes:

- I would not describe the exposure as fully closed. This removes the unconditional shell disclosure, which is the defect that was reported, but git's own transport and authentication diagnostics still inherit stdout and stderr unredacted, and one connection-failure case does not cover all of them.
- Once the clone succeeds, an embedded token is persisted in `/kb-main/.git/config` inside the volume. That follows from cloning with credentials in the URL rather than from anything this change does wrong, and provisioning the credential separately is the way out of it. Out of scope here.

On `deploy/k8s/statefulset.yaml`: agreed, leave it, and thank you for offering. I will file it separately rather than widen this PR. My framing of why it is low risk would have been wrong, though, and it is worth correcting in public: mounting an SSH key does not constrain what that secret actually holds, so an HTTPS credential configured there does leak through the initContainer's echo today. It is pre-existing and it is mine to fix, not yours.

Last thing: `main` has moved since `380d6cb`. Of the four files you touch, only `CHANGELOG.md` differs, and it now carries an empty `[Unreleased]` block above `## [0.7.2] - 2026-09-07`, so a rebase that keeps your entries under `[Unreleased]` should be mechanical. The checks on this fork branch still need a maintainer approval before they run, so none of the above is a statement about CI.

yuzi-co:
You are right about the gap, and I reproduced it before fixing it. With the test file as you reviewed it (`3ce063a`), mutating the line to `echo "[entrypoint] /kb-main is empty; cloning ${KB_GIT_REMOTE_URL}"` passes all four tests, for exactly the reason you give: the case that reads the output never sets that variable, and the case that sets it never reads the output.

Fixed in `2469aef`. The credential case is now parametrized over both variables, so each one gets a run where it carries the token and the output is checked. I also assert the emitted output contains no `://` at all, which covers URL shapes the two cases do not spell out.

Re-ran the mutation harness against the current branch:

| mutation | result |
|---|---|
| `cloning ${clone_url}` | caught (2 tests) |
| `cloning ${KB_GIT_REMOTE_URL}` | caught (1) |
| `cloning ${KB_REMOTE_URL}` | caught (1) |
| fallback removed | caught (2) |
| precedence inverted | caught (1) |
| same `${KB_GIT_REMOTE_URL}` mutation against `3ce063a` | not caught |

The last row is your finding, kept in the table so the gap is on record rather than quietly closed.

## On the wording

Both corrections taken, and the overclaim was mine.

The CHANGELOG entry and the PR description now say this removes the unconditional shell disclosure rather than closing the exposure: git inherits the same stdout and stderr, one connection-failure observation is not a survey of its diagnostics, and a token in the clone URL is written to `/kb-main/.git/config` once the clone succeeds. Both of those follow from having a credential in the URL at all, and provisioning it separately is the way out. Neither is something this PR fixes.

The k8s claim is likewise narrowed to what it can support: remote selection is unchanged there because both variables bind to the same secret key, not that every behaviour is. If that path ever did bootstrap, it would log a different line.

Noted on `statefulset.yaml`, and on the reason. Leaving it with you.

## Rebase

Rebased onto `a1913ad`. `CHANGELOG.md` was the only conflict; both entries now sit under the current empty `[Unreleased]`, above `## [0.7.2] - 2026-09-07`. The other three files applied unchanged.

Local run on `2469aef` from a fresh clone, installed as CI does with `uv pip install -e '.[dev]'`: `sh -n` clean, `ruff check .` clean, `mypy src` clean, `pytest` exit 0 with the same 4 environment skips. Still nothing verified on your side until the fork run is approved.

ajdelaguila:
`2469aef` closes it. Reviewed at that head; the earlier verdict does not carry over.

The parametrized credential case is the right shape. Restoring an echo of either
variable now fails its own case, and keeping the failing row for `3ce063a` in the
mutation table is the part I appreciate most: a coverage gap that is recorded is
worth more than one that is quietly closed.

Checks on this head are green: `test`, `test-python-3-14`, `changelog-guard`,
`doc-consistency-guard`, `version-free-guard`, `lint-title`. Note for the record
that no CodeQL or `Analyze` check appeared on this head; that analysis runs
against the default branch, so this change gets its coverage after the merge
rather than here.

Two observations from the review at this head. Neither is a change request, and
neither should hold the PR up:

- Exhaustive echo-regression coverage still is not established, which is a
  different claim from the one you fixed. A conditional form such as
  `echo "... ${KB_REMOTE_URL:+${KB_GIT_REMOTE_URL}}"` discloses only when both
  variables are set, and each credential case sets exactly one while the
  both-set precedence case discards output. Closing that would mean a case with
  two distinct credentials whose output is checked. I mention it so the ceiling
  of the current suite is on record, not because I think this PR should grow.
- `assert "://" not in out` is wider than the property it protects. It applies
  to the whole block's combined output, so a future bootstrap message that
  linked to public troubleshooting docs would fail it for no security reason.
  The token and host assertions already catch direct disclosure. If it ever
  becomes annoying, narrowing it to the clone announcement is the fix, and that
  is maintenance on our side rather than something you owe.

The rebase is clean. `entrypoint.sh` and `compose.yaml` are byte-identical to
what I reviewed at `9a721af`, blob identifiers included, and the test delta is
confined to the credential case.

This is ready to merge from my side. One process note, since it affects timing
rather than the code: `main` moved after your rebase, and including this in the
next release means the release revision is re-prepared and re-reviewed on top of
your merge rather than published from the commit that preceded it. That is our
work, not yours, and nothing further is needed from you.

Thanks for the care on this one, particularly for narrowing your own wording in
the CHANGELOG rather than leaving the stronger claim standing.

ajdelaguila (approved):

Issue #111, opened 2026-07-08
Research: publish MCP server metadata for discovery and catalog ingestion


ajdelaguila:
## Source signal

The outreach pass surfaced OpenHands/extensions#351, which is adding generated MCP `server.json` artifacts from integration catalog entries:

- https://github.com/OpenHands/extensions/issues/351
- Data Olympus field note: https://github.com/OpenHands/extensions/issues/351#issuecomment-4911209847

The same need appears in the outreach index for the official MCP Registry, `punkpeye/awesome-mcp-servers`, Smithery, PulseMCP, MCP Directory, and Copilot-facing resource submissions.

## Current Data Olympus behavior

Data Olympus exposes a Streamable HTTP MCP server and has setup documentation, but it does not yet publish a stable machine-readable server metadata artifact for registries or downstream catalogs.

That leaves discovery targets with incomplete information about:

- transport and default local endpoint;
- self-hosted versus SaaS assumptions;
- required local git-backed KB bundle path and index path;
- optional remote/push configuration;
- read-only lookup versus proposal/write capability;
- license, categories, package/install hints, and security expectations.

## Research questions

- Which metadata format should be canonical for Data Olympus: official MCP `server.json`, repository-local catalog metadata, or both generated from one source?
- How should local-first configuration be represented without implying a hosted SaaS endpoint?
- Should write/proposal capability be first-class metadata so catalogs and agent hosts can distinguish retrieval-only use from governed KB mutation?
- What minimum metadata is required for official MCP Registry publication and high-visibility MCP directories?
- Can the same metadata support Copilot/Codex/Claude Code-facing packaging without duplicating install instructions?

## Candidate shape

Evaluate a single source of truth that can generate or validate:

- `server.json` or equivalent official MCP registry artifact;
- directory/listing snippets for MCP directories;
- README install metadata;
- future Codex/Copilot/Claude Code packaging hints.

Candidate Data Olympus-specific fields may need to describe:

- `kb_bundle_path` or workspace-local knowledge bundle path;
- `index_path` / generated index location;
- optional git remote and push behavior;
- allowed modes: lookup, consult gate, propose memory, resolve pending;
- deployment mode: local, team-hosted, or behind auth/proxy.

## Evaluation criteria

- A user can discover Data Olympus through MCP catalogs without reading a long issue comment.
- Metadata does not misrepresent Data Olympus as a hosted SaaS integration.
- Generated artifacts remain consistent with setup documentation.
[...]

PR #260, opened 2026-09-12
feat(registry): declare server.json and the mcp-name marker for the official MCP registry

ajdelaguila:
## Why

The official MCP registry is where `modelcontextprotocol/servers` now sends new server implementations: its CONTRIBUTING refuses them outright, and its README tells readers looking for a list of servers to browse the registry instead. data-olympus is not published there. Two searches of the registry's v0 API, `?search=data-olympus` and `?search=knaisoma`, returned zero servers on 2026-09-12.

Context and the research that led here: #111.

## What this adds

**`server.json`** at the repository root, validated against the published schema with `jsonschema`:
- `name` is `io.github.knaisoma/data-olympus`. The `io.github.*` namespace authenticates through GitHub, so no DNS verification is needed.
- The package entry is `registryType: pypi` against `https://pypi.org`, which is on the registry's accepted-source list.
- The transport is `streamable-http` at `http://localhost:8080/mcp`, which is what `data-olympus-mcp` actually serves and what `docs/adoption.md` documents. **Not stdio.**

**The ownership marker** in `README.md`: `<!-- mcp-name: io.github.knaisoma/data-olympus -->`. The registry verifies PyPI package ownership by finding `mcp-name: $SERVER_NAME` in the README that PyPI serves as the package description.

**`docs/mcp-registry.md`**, following the `docs/glama.md` precedent: what the file declares, what publication still needs, and a checklist.

## What this deliberately does not do

- **It does not publish.** Publication authenticates as a maintainer through `mcp-publisher login github`, or from Actions through OIDC. That is a deliberate step, not a file change.
- **The marker is not on PyPI yet.** The published 0.7.3 description lacks it; the marker takes effect with the next release. The checklist covers updating **both** version fields and confirming the marker on that exact release.
- **No runtime launch metadata.** The package is `data-olympus` but the MCP console script is `data-olympus-mcp`, so `uvx data-olympus` would start the CLI. The explicit form is documented. Whether to declare automated launch is a separate decision that should be tested against a real client rather than inferred from a schema that validates.

## Verification

- `uv run pytest`: green, with two skips for optional dependencies (`sentence_transformers`, `tiktoken`). Note this needs CONTRIBUTING's documented setup (`uv pip install -e '.[dev]'`); a plain `uv run pytest` in a fresh worktree syncs a non-editable build and imports a stale `data_olympus` 0.5.0, which fails three tests on unmodified `main` too.
- `uv run ruff check .`: all checks passed.
- `uv run data-olympus lint example-bundle`: `0 errors across 0 files (14 linted)`.
- `server.json` validated against `https://static.modelcontextprotocol.io/schemas/2025-12-11/server.schema.json`.
- The repository's own pre-commit enforcement gate passed on each commit.
- No path touched triggers the changelog guard (`src/`, `bin/`, `deploy/`, `SPEC.md`), and no OKF schema change is involved, so no Spec Proposal applies.

## Companion review (Codex)

Three rounds. Round 1 rejected an unsupported claim that several catalogues treat a registry entry as a credential, and a checklist that updated only one of the two version fields. It also caught the `data-olympus` versus `data-olympus-mcp` executable gap. Round 2 rejected the verification command: it read the latest-release endpoint rather than the pinned one and counted the fragment `mcp-name` anywhere in the response, so a newer marked release could have masked an unmarked pinned one. The corrected command was run as written and returns `false` for 0.7.3, as it should. Round 3 approved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 [![knaisoma/data-olympus MCP server](https://glama.ai/mcp/servers/knaisoma/data-olympus/badges/score.svg)](https://glama.ai/mcp/servers/knaisoma/data-olympus)
 [![MCP Marketplace](https://img.shields.io/badge/MCP%20Marketplace-Indexed-blueviolet)](https://getlulu.dev/mcps)
 
+<!-- mcp-name: io.github.knaisoma/data-olympus -->
+
 **New here? Start with [WHY.md](WHY.md).** It is the story behind the project: the
 problem we kept hitting with coding agents, what data-olympus does differently, how
 it relates to Google's Open Knowledge Format, and where our benchmarks say it is
@@ -65,6 +67,7 @@ See `docs/adoption.md` for the full bundle authoring guide.
 - [`docs/comparison.md`](docs/comparison.md): how data-olympus relates to OKF, enterprise catalogs, markdown KB tools, agent-context conventions, RAG, and ADR tooling.
 - [`docs/okf-profile.md`](docs/okf-profile.md): field-by-field OKF profile — which governance extensions are stable, which are runtime-only serving fields, and which are experimental candidates.
 - [`docs/glama.md`](docs/glama.md): Glama registry claim, release, and score-maintenance notes.
+- [`docs/mcp-registry.md`](docs/mcp-registry.md): official MCP Registry notes, what `server.json` declares, and the checklist for publishing.
 - [`docs/enforcement.md`](docs/enforcement.md): turning the KB into a mandatory consultation gate (hooks, `kb enforce`).
 - [`benchmarks/README.md`](benchmarks/README.md): retrieval benchmark methodology and how to reproduce the numbers in `docs/comparison.md`.
 - [`SECURITY.md`](SECURITY.md): supported versions and how to report a vulnerability.
--- a/docs/mcp-registry.md
+++ b/docs/mcp-registry.md
@@ -11,8 +11,7 @@ research task. It is the registry equivalent of [`glama.md`](./glama.md).
 
 `modelcontextprotocol/servers` no longer accepts new server implementations. Its
 CONTRIBUTING directs authors to the registry, and its README tells readers looking for
-a list of servers to browse the registry rather than that repository. Several catalogues
-treat a registry entry as a credential.
+a list of servers to browse the registry rather than that repository.
 
 ## What is in this repository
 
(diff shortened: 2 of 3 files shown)


ajdelaguila:
Incorporated into the 0.8.0 release PR #258 rather than merged separately, so the release carries it in one merge.

All three commits here were cherry-picked onto the release branch from the same `dd8bce6` base with no conflicts, preserving authorship. `server.json` and `docs/mcp-registry.md` are byte-identical to this branch's head `6a996d0`; `README.md` necessarily differs only because the release branch had already changed that file.

Two changes were made on top, both worth flagging here since they alter what this PR proposed.

**Both `server.json` version fields are now `0.8.0`, not `0.7.3`.** Correct here, where the change sits against `main`, and wrong in a release: the registry resolves `packages[0].version` against PyPI and reads the `mcp-name` marker out of the description served for *that* version, so shipping `0.7.3` would have pointed the entry at a description predating the marker. This PR's own checklist already called for the update; it just had to happen as part of the release rather than after it.

**`tests/test_server_json.py` is new.** Nothing covered `server.json` at all: `version-free-guard` only reads `[project].version` from `pyproject.toml`, `doc-consistency-guard` only the SPEC/adoption enums, and no test or workflow referenced the file. The tests assert both version fields track the declared release, that the README marker names exactly the server in `server.json`, that the package identifier is the distribution actually published, and that neither version is a placeholder.

The publication notes were also reworded so they stay true after the release lands rather than describing a state that 0.8.0 ends, and the changelog and release notes now say 0.8.0 is the first *stable* release carrying the marker, since candidates publish to PyPI first with the same README.

Verified at the release head: the built wheel `METADATA` and sdist `PKG-INFO` both report `Version: 0.8.0` and contain the exact marker string, so the description PyPI serves for 0.8.0 will carry it, and `server.json` validates against the published schema. `server.json` is deliberately not included in the sdist: `mcp-publisher` reads it from the repository root, and ownership verification reads the marker from the description.

Leaving this open and the branch untouched until #258 merges, then closing it as incorporated. Nothing here is lost if that merge does not happen.

ajdelaguila:
Closing as incorporated. These commits were cherry-picked into #258, which is merged to `main` and ships in 0.8.0.

Compared against `main` at 457ad7a:

- `server.json` matches this PR except for both version fields, which were pinned to 0.8.0. The registry resolves `packages[0].version` against PyPI and reads the ownership marker from that version's description, and 0.8.0 is the first stable release carrying the marker. `tests/test_server_json.py` guards that pin.
- The `mcp-name` marker is in `README.md`.
- `docs/mcp-registry.md` includes corrections made during review so the publication notes stay accurate after the release lands.

As before, this declares the registry entry but doesn't publish it. Publication is a separate maintainer step, described in `docs/mcp-registry.md`.

Issue #270, opened 2026-09-15
Health lists the pending id of every path lock to unauthenticated callers


ajdelaguila:
## Problem

Since #253, the health response carries `path_locks`, one record per held path lock with `target_path`, `owner_kind`, `pending_id`, `acquired_at` and `age_seconds`. Compact health includes the list whenever it is non-empty, and neither `GET /api/v1/health` nor MCP `kb_health` requires authentication, even when auth is configured. The same model is returned as the body of the degraded 503 that REST read routes send.

So an unauthenticated caller can enumerate the pending ids and target paths recorded in readable pending-proposal locks, including claimed proposals and orphaned locks. Auto-commit locks also expose their owner marker in the `pending_id` field.

The impact is bounded. When auth is configured, listing or reading pending entries (`kb_list_pending`, `kb_get_pending`, `GET /api/v1/pending`) requires an authenticated caller, and resolving one requires the resolve capability, so a leaked id does not expose a postimage or allow approval on its own. It does contradict the rest of the auth model, where pending-queue metadata is only available to authenticated callers, and it undermines any read surface that tries to show pending ids to authenticated callers only, which #241 is about to add.

Found while writing the acceptance checks for #241.

## Proposed fix

When auth is configured, health lock records for a caller without a valid token carry only `target_path`, `owner_kind`, `acquired_at` and `age_seconds`. A valid token adds `pending_id`. With auth not configured nothing changes. The operator use case from #253, finding which path is wedged, needs only the path and the age.

## Acceptance criteria

- With auth configured, compact and verbose health over REST and MCP, and the degraded 503 body on REST read routes, omit `pending_id` for a missing or invalid token and include it for a valid one.
- A missing or invalid token does not change the health status code.
- With auth not configured, health output is unchanged.
- When auth is configured, health responses carry `Cache-Control: private` and `Vary: Authorization`.
- No lock field added later (for example the contest fields planned in #241) appears in health unless this issue's rule is extended to it explicitly.

ajdelaguila:
Fixed on `main` by #279 (`69e94d6`), not yet in a released version.

With auth configured, `GET /api/v1/health`, the degraded 503 body that every read route can return, and MCP `kb_health` now omit each `path_locks` record's `pending_id` for a missing or invalid token; a valid token still sees it. Behaviour is unchanged when auth is not configured. Responses served with auth configured also carry `Cache-Control: private` and `Vary: Authorization`.

[...]

Issue #272, opened 2026-09-15
Path lock publication: a reader can see a half-written lock, and a failed directory sync leaves a pending entry without its lock


ajdelaguila:
## Problem

`PendingQueue._acquire_lock` creates the lock with `os.open(..., O_CREAT | O_EXCL | O_WRONLY)` and then writes the JSON into that descriptor. Between the create and the end of the write, the lock file exists but is empty or truncated. Writers are serialized, but readers do not take the serializer, so a reader can observe that state.

Reproduced at `5bdcfd4` by creating a lock file without writing its contents: `derive_running_contest` reports `under_review: False`, and `held_locks()` omits the record. `locks_held()` still counts the file, so health can report a held lock without identifying its path. This simulates the intermediate state; it does not measure how often the race occurs.

The window is short and nothing here corrupts data, since the exclusive create still serializes acquisition. The effect is on observers: diagnostics and any read surface built on locks can report a held path as free. #241 plans to put contest ids in the lock, which makes the write larger and the lock more important to read correctly.

Found while writing the acceptance checks for #241.

### A related failure on the same path

`enqueue` writes the lock, then the entry through `atomic_write_json`, and releases the lock if the entry write raises. `atomic_write_json` renames the entry into place before it fsyncs the parent directory, so a failure in that fsync raises after the entry is already published. Reproduced at `5bdcfd4` by making the directory fsync fail: `enqueue` raises, the entry file remains, and its lock is gone, so another proposal can take the path while the first is still pending.

## Proposed fix

Publish a complete lock atomically while keeping exclusive-create semantics, for example by writing a temporary file in the lock directory and `os.link`-ing it onto the lock path, which fails if the lock already exists. Temporary files must not end in `.lock`. Separately, a failure after the entry is published must not release its lock: either keep the lock and surface the entry as needing reconciliation, or remove the published entry before releasing, whichever the recovery path in #253 handles better.

## Acceptance criteria

- A reader never observes an existing lock file with empty or partial content, tested by pausing a writer between creating and publishing.
- Concurrent acquisition of one path still has exactly one winner, the existing ownership fields and `0o600` permissions are preserved, and failure cleanup never removes another owner's lock.
- Temporary files are ignored by lock enumeration and removed after handled failures. Orphan-lock GC recognizes the temporary naming scheme and reaps crash leftovers past an age threshold, with creation, publication and reaping under the same serializer so cleanup cannot remove a file mid-publication.
[...]

Issue #112, opened 2026-07-08
Write path: governed-lane protection (agents can propose, only humans promote)


ajdelaguila:
## Summary

Agent-lane writes must never be able to mint or silently rewrite governing rules. Three mechanisms, shipped together as "agents can propose, only humans can promote":

- **Status clamp**: a non-operator-confirmed write (auto-commit at any confidence) may not set or change `status` INTO the in-force class (`active`/`accepted`/`approved`). Such a write is demoted to a pending entry, never rejected.
- **Governed-target edit demotion**: any agent-proposed edit whose target document is currently in-force never auto-commits, regardless of confidence; it always lands as a pending entry for operator review.
- **Injection-pattern annotation**: a heuristic scan of postimages (agent-directed imperatives, "ignore previous instructions", exfiltration URLs, encoded blobs) that annotates the pending entry for the reviewer. Advisory only, never a hard gate.

## Motivation

Forged authority is the residual hole after the authority research (#109): under allow-by-default, a high-confidence `kb_propose_edit` can today commit a postimage claiming `status: accepted` straight into an indexed prefix, and an edit to an already-accepted doc changes governing content while keeping its accepted status. Rules spread to the whole team via git pull + index refresh, so one poisoned write propagates fast. With this control, forging a governing rule requires getting a human to approve it in pending review, where provenance (agent identity, session, confidence, commit trailers, tamper-evident audit chain) is already in front of them.

Explicit non-goals: protection against a malicious human reviewer, or against pushes to the git remote outside the MCP (branch protection and commit signing are ops-level controls, documented in SECURITY.md).

## Behavior and rollout

- Enabled by default; env opt-out `KB_GOVERNED_LANE_PROTECTION=off` restores current behavior exactly.
- Not a breaking change: no write is rejected; affected writes land as pending instead of committed.
- Feedback loop so demotions are never silent:
  - per-write: the tool response carries an explicit notice plus an instruction to the model to inform the operator
  - per-session: a recap surface (audit log filtered by `source_session`: N committed, M pending) exposed as a tool/endpoint and included in the `kb_consult`/`kb_health` pending-actions envelope
  - optional SessionEnd/Stop hook in the agent-hooks installers that prints the recap
- CHANGELOG flags the default-behavior change prominently.

## Relationships

- Complements #71 (secret scan) - same content-gate family on the write path.
- Follows from the #109 decision (authority envelope research); the forgery risk was explicitly split out of that slice.

Part of the v0.4.0 release program.

PR #278, opened 2026-09-17
feat(write): add contest input validation and lock persistence (#241)

RemanenetSpy:
﻿### Summary
This PR implements PR 1 (write-side validation and lock persistence) for issue #241, establishing write-time governance for proposals that challenge or contradict existing documents without leaking internal path structure or ghost states.

Per maintainer coordination in #241, this PR lands the write-side changes against upstream `v0.8.2` prior to PR 2 (read-side 5-field projection) and upcoming atomic publication changes (#272).

### Technical Changes
1. **Two-Phase Write Validation (`src/data_olympus/tools_write.py`):**
   - **Stage 1 (Pre-Rate-Limiter):** Validates payload shape, allowed fields (`"contradicts"`, `"reason"`), array bounds (1–10 items, 1–200 chars each), contest reason bounds (<= 500 chars), and enforces secret scanning precedence returning `rejected_secret_detected` (HTTP 422) or `rejected_invalid_contest` (HTTP 400).
   - **Stage 2 (Post-Rate-Limiter):** Validates contradiction targets against `idx.id_to_path_map()`. Prevents target self-contradiction (`target_path in contradicts`), rejects non-existent document IDs without reflecting keys in the response (zero-leak invariant), and returns `rejected_contest_index_unavailable` (HTTP 503) if the index snapshot is missing.
2. **Lock Lifecycle Retention (`src/data_olympus/pending.py`):**
   - `_acquire_lock`: Serializes `intent: "contest"` and `contradicts` into `.lock` JSON payloads.
   - `_claim` & `restore_resolve`: Preserves `intent` and `contradicts` across entry claim transitions and recovery aborts.
3. **Dual Rationale Projection (`src/data_olympus/models.py`, `pending.py`, `tools_write.py`):**
   - `PendingEntry` and `kb_list_pending` surface proposal `reason` and `contest.contest_reason` as distinct fields.
4. **REST API & Server Schemas (`src/data_olympus/rest_api.py`, `server.py`):**
   - `propose_edit` forwards `contest` payload to `kb_propose_edit_fn`.
   - `propose_memory` explicitly rejects `contest` payloads with HTTP 400.
   - `_propose_status` maps `rejected_contest_index_unavailable` -> 503, `rejected_invalid_contest` -> 400.
   - FastMCP server tool accepts annotated `contest: ContestParam = None`.

### Verification & Tests
- **Test Suite (`tests/test_contest_validation.py`):** 17 comprehensive unit and integration tests covering:
  - Malformed payload rejection (non-dict, unknown keys, empty/whitespace strings, out-of-range lengths).
  - Secret scanning precedence on `contest.reason` and `contest.contradicts`.
  - Target self-contradiction prevention and non-existent ID rejection.
  - Index unavailable 503 status mapping.
  - Lock persistence across `_claim` and `restore_resolve` transitions.
  - Dual rationale separation on `kb_list_pending`.
  - REST API endpoint HTTP 400 and 503 status codes.
- **Regression Gates:**
  - `pytest tests/test_contest_validation.py tests/test_running_contest.py`: 28/28 passed in 1.96s.
  - `ruff check .`: 0 errors across entire repository.


ajdelaguila:
Thanks for turning this around quickly, and noted on the ordering from #241: this one lands first, and the atomic publication change in #272 rebases over it.

Checks are green on `12d2c25` (`test`, `test-python-3-14`, the three guards, and the dry-run build). A full technical review of the diff follows; expect it by 2026-09-19.

One sequencing point is worth raising ahead of that review, because it decides what changes and in which pull request.

`held_locks()` here adds `intent` and `contradicts` to every lock record. That list is exactly what `_build_health` passes through as `path_locks`, and per #270 the health response is still served to callers without a token on `main` today. So on `main`, those two fields would be readable unauthenticated until #270 is fixed.

The fix for #270 is written as an allow-list over lock-record fields rather than a `pending_id` special case (`target_path`, `owner_kind`, `acquired_at`, `age_seconds`, plus `pending_id` for an authenticated caller). Two consequences for this PR:

- Before that fix merges, `intent` and `contradicts` are visible to an unauthenticated caller.
- After it merges, both fields are dropped from `path_locks` for *every* caller, authenticated included, until the allow-list is extended to them deliberately. The projection added here would go quiet rather than stay authenticated-only.

Writing `intent` and `contradicts` into the lock file itself is unaffected. That is the part agreed in #241, claim and restore keep working off it, and it stays either way.

Two ways to sequence it:

1. Keep the lock payload change and drop the `held_locks()` projection from this PR. The allow-list entry and the projection then land together once #270 is fixed, and there is no window and no dead output.
2. Keep it as written. The exposure window is bounded by the gap between the two merges, and the allow-list has to be extended in the same change that closes #270 or the fields stop appearing.

Either works from here, so it is your call. 1 is the smaller diff and is the one I would pick.

RemanenetSpy:
@ajdelaguila Implemented Option 1. Dropped the \intent\ and \contradicts\ projection from \held_locks()\ to eliminate unauthenticated exposure prior to #270 merge, keeping on-disk lock persistence (\_acquire_lock\, \_claim\, and \
estore_resolve\) and contest validation intact. Tests updated to verify lock file attributes on disk directly.

ajdelaguila:
Option 1 received, and it is exactly what landed: I diffed `8885c57` against `12d2c25` and the only changes are the removal of the two `intent` / `contradicts` copies in `held_locks()` and the three assertions in the lock-persistence tests that now read the `.lock` JSON from disk instead of the projection. `_acquire_lock`, `_claim`, `restore_resolve`, the validation stages, the REST and FastMCP surfaces and the CHANGELOG entry are unchanged.

That closes the sequencing question in both directions:

- Nothing unauthenticated can see `intent` or `contradicts` before #270 is fixed, because they no longer reach `path_locks` at all.
- The #270 fix needs no allow-list extension and produces no dead output, so neither change has to wait on the other.
- The part agreed in #241, persisting both fields in the lock file across claim and restore, is intact and now asserted against the file itself, which is a stricter check than the one it replaced.

All checks are green on `8885c57` (`test`, `test-python-3-14`, `changelog-guard`, `doc-consistency-guard`, `version-free-guard`, `lint-title`, `dry-run / build`).

On the #272 ordering: I measured the overlap rather than estimating it. The two changes touch the same lines of `_acquire_lock`, and the resolution is about six lines confined to that one function, with the signature and docstring merging cleanly. That cost is mine to pay, not yours, and it does not change anything about this PR.

Nothing is needed from you at this point. The full technical review of the diff follows on 2026-09-19 as previously stated.

ajdelaguila:
Full review of `8885c57`, as promised. Base `main` @ `54252f18`, +784/-24 over 7 files, all checks green. I read the diff and the surrounding source; I did not run your branch locally.

Overall: this is solid work. The two-stage split (in-memory validation before the rate limiter, index resolution after it) is the right shape, and the lock-persistence tests assert against the file on disk, which is stricter than what they replaced. The zero-leak discipline on the rejection paths is real rather than claimed: no submitted ID or reason value reaches a contest-validation rejection response or its audit event. (The pending listing is a separate, deliberate surface, and it does publish `contradicts` and `contest_reason` — that is the design, not a leak.) Two questions I had going in are closed and need nothing from you:

- **No bypass on the MCP path.** `kb_propose_memory` in `server.py` takes no `contest` parameter and no passthrough, and the search-mode `call_tool` proxy re-enters the validated chain for the inner tool, so the memory path is closed by signature as well as by the REST body check you added.
- **`bin/kb` is correctly untouched.** `cmd_propose_edit` does not send `evidence` either, so a server-side field with no CLI flag is the existing precedent, not a gap this PR introduces.

Points below are tagged so the required set is unambiguous: **1 and 2 are must-fix before merge**, 3 and 4 are should-fix or tell me why not, 5 and 6 are worth doing, 7 is an open question rather than a defect, and the rest are nits you can take or leave.

### 1. A failed index query is reported as invalid client input, not as the promised 503 — must fix

This is the one I would most want fixed, and it is partly our fault rather than yours.

`_check_contest_index` wraps `idx.id_to_path_map()` in `except Exception` and maps a raised error to `rejected_contest_index_unavailable` / 503. But `id_to_path_map()` does not raise on a query failure. `index.py:2225` catches `sqlite3.Error` around the `SELECT id, path FROM docs` and returns `{}`:

```python
conn = self._connect()
try:
    rows = conn.execute("SELECT id, path FROM docs").fetchall()
except sqlite3.Error:
    return {}
finally:
    conn.close()
```

So when the connection opens but the query fails — a missing `docs` table, database corruption, a locked table — the mapping comes back empty, the first submitted ID looks absent, and the caller gets `rejected_invalid_contest` and **HTTP 400**. The description promises 503 "if the index snapshot is missing", and a client that trusts the status code will treat a server-side outage as its own malformed input and not retry. `test_check_contest_index_unavailable` raises `RuntimeError` from the mock, which cannot reach this path, so nothing currently catches it.

The fix needs an index read that distinguishes "query failed" from "successfully read, no rows" — the empty index is a legitimate state on a fresh KB and must stay a 400-class answer, so please do not classify every empty mapping as unavailable. If it is easier to change `id_to_path_map` to let `sqlite3.Error` propagate, say so and I will take that change on our side instead; the existing callers are the write-path validation gate and governed-lane classification, and I would rather fix the shared contract once than have contest work around it. Either way it wants coverage for both cases: query failure and genuinely empty index.

### 2. `demotion_reason = "contest_declared"` changes behaviour today, not just the contract — must fix

I had this down as a documented-contract break. Looking again at the branch it selects, it is a live bug.

`models.py:438-446` documents `ProposeResponse.demotion_reason` as governed-lane write protection under `KB_GOVERNED_LANE_PROTECTION` (issue #112), enumerates its three permitted values (`status_promotion`, `governed_target`, `governed_target_unverified`), and states it is `None` on every other outcome. `PendingEntry.demotion_reason` and `AuditEvent.demotion_reason` point at that same definition.

`_park(demotion_reason or ("contest_declared" if clean_contest is not None else None))` makes `park_demotion_reason` non-null for a clean contest. That selects the branch at `tools_write.py:1590`, which returns the operator prompt beginning "was DEMOTED to pending review by governed-lane write protection (reason: contest_declared)". A plain contest park — no lane protection involved, possibly with the feature disabled entirely — now tells the operator that lane protection demoted it. It also skips the ordinary branch below, so the response loses `proposal_text=postimage`.

The narrow fix is to restore `return _park(demotion_reason)`, removing only the `"contest_declared"` fallback. Please keep it that way round rather than forcing `None` for a contest park: a proposal can be both governed-lane-demoted *and* a contest, and that genuine `demotion_reason` must survive. `meta["intent"] = "contest"` is the honest signal and `kb_list_pending` already projects it; if the response needs to say it out loud, a separate field is the place. `test_kb_propose_edit_contest_parks_and_persists_lock` asserts `resp.demotion_reason == "contest_declared"`, so that assertion moves with the fix, and it is worth asserting the returned prompt is the ordinary one.

### 3. The test suite and description claim REST route coverage that does not exist — should fix

Under "Verification & Tests", the description lists "REST API endpoint HTTP 400 and 503 status codes" among what `tests/test_contest_validation.py` covers, and the module docstring makes the same promise under heading 6. The only REST test in the file is `test_rest_api_propose_status_mappings`, which makes four assertions against the pure `_propose_status` helper. The branch you added in `rest_api.py::propose_memory` — the `if "contest" in body` early return — is never executed by any test.

To be clear about the severity: this is a documentation-truthfulness problem, not a demonstrated bypass. I am not claiming the branch is wrong, only that nothing exercises it. Either resolution is fine: add a route-level test through the ASGI app, or cut the coverage claim from both the docstring and the description. A stated guarantee with nothing behind it is the only thing that cannot stay. Given that branch is what keeps a contest payload off the memory path, I would lean toward the test.

### 4. `_check_contest_index` reports internal defects as a transient index outage — should fix

```python
try:
    id_path_map = idx.id_to_path_map()
    target_ids = {...}
    for cid in contradicts:
        ...
except Exception:
    return ProposeResponse(status="rejected_contest_index_unavailable", ...)
```

The `except Exception` spans the index access *and* the whole validation loop. A genuine bug inside the loop surfaces to the caller as `rejected_contest_index_unavailable` and, at REST, as a 503 that invites a retry which cannot succeed. Narrow the `try` to the `id_to_path_map()` call and let the loop's own errors surface. Note this is about diagnosis, not about item 1 — narrowing the `try` does not fix the misclassification there, because nothing raises in the first place.

### 5. Pin the zero-leak invariant where it is strongest

`test_check_contest_index_doc_not_found` checks `"STD-999" not in err.reason`: one substring, one field, one response. The invariant worth pinning is that a submitted ID never reaches a contest *rejection* response or its audit event — which is the whole serialized `ProposeResponse` plus the emitted audit record, not just `reason`. It looks clean by construction (`_emit_audit` passes the static error reason on both contest paths), but nothing holds it there. Asserting the submitted ID is absent from `resp.model_dump_json()` and from the audit record would make that claim load-bearing.

### 6. State the deferred read-side projection somewhere durable

Nothing in the description is false after the Option 1 change, but it still reads as though the read surface were in scope. A reader cannot tell that leaving `intent` and `contradicts` out of `held_locks()` is deliberate and sequenced behind #270. One line in the description or the CHANGELOG entry saying the read-side projection is deferred, and why, prevents someone re-adding it later without knowing what it was waiting for.

### 7. Secret-flagged `contest.reason` is rejected where the neighbouring fields redact — open question

Not a defect, and I am not asking you to change it; I want the choice on the record.

`_validate_contest` returns `rejected_secret_detected` when a secret pattern hits `contest.reason`. The two adjacent advisory fields do the opposite: `tools_write.py:1432-1442` redacts a flagged `reason` in place and lets the write proceed, with the rationale recorded inline from issue #71 ("reason is advisory metadata, not committed content, so redaction preserves the operation while keeping every downstream surface clean"), and `_redact_evidence` gives `evidence` the same treatment. `contest.reason` lands in pending meta and the pending listing, never in committed content, so the redaction argument reaches it too.

Refusing to accept credential-bearing input is a perfectly defensible policy, and `rejected_secret_detected` describes what actually happened. It is the inconsistency across one class of field I would like decided deliberately: either contest joins the redact-and-proceed pattern, or a line somewhere says contest input is held to a stricter rule and why.

### Nits

- **Redundant disjunct.** `if id_path_map[cid] == target_path or cid in target_ids`: `cid in target_ids` is only true when `id_path_map[cid] == target_path`, so the second test cannot fire on its own, and `target_ids` costs a full pass over the map on every contest write. No test reaches that branch, which is consistent with it being unreachable.
- **Projection keyed on a field rather than the intent.** `entry["meta"].get("intent") == "contest" or entry["meta"].get("contradicts")` in `pending.list()` emits a `contest` object for any entry carrying `contradicts`, whatever its intent. `meta` is built internally today so it is not reachable, but the `or` makes the projection depend on data instead of the declared intent. Untested branch.
- **Undocumented rule.** Duplicate `contradicts` items are rejected with `rejected_invalid_contest`. Good rule, worth a line in the description or CHANGELOG.
- **The Windows shim is untested by this CI.** `_scoped_win_atomic_writes` is autouse but gated on `sys.platform == "win32"`, and CI runs Linux-only on both Python versions, so nothing here exercises it. Worth knowing that it patches `atomic_write_json` and `atomic_remove` in both `pending` and `durable`, which is the write path #272 rewrites; if Windows CI is ever added, those two will need reconciling.
- **Two private names in the rewritten tests.** The lock-persistence tests now build the lock path from `_locks_dir` (`pending.py:152`) and `_path_lock_filename` (`pending.py:115`). Both are correct today and Option 1 leaves little alternative, but a one-line comment saying the coupling is deliberate would help whoever changes that filename scheme.
- **Leftover intermediate.** The projection removal left `rec: dict[str, Any] = {...}` plus `out.append(rec)` where the original had a direct `out.append({...})`. Reverting those two lines makes this hunk's diff empty, which is the honest representation of the change.

### On #272

Nothing for you here. I measured the overlap with `git merge-tree`: `models.py`, `rest_api.py` and `server.py` auto-merge clean, `_acquire_lock`'s signature and docstring auto-merge, and the only conflicts are a mechanical CHANGELOG merge and about six lines in one function where your two optional fields meet the atomic publication sequence. That is mine to resolve whichever way the merges land.

Items 1 and 2 are what I would want fixed before this merges. Items 3 and 4 I would want either fixed or answered. The rest can land here or as follow-ups, your call.

RemanenetSpy:
@ajdelaguila All review items have been addressed in commit `5e43290`:

1. **Item 1 (Must-fix):** In `src/data_olympus/index.py::id_to_path_map`, removed the broad `try/except sqlite3.Error` so database errors propagate cleanly. `_check_contest_index` now isolates `idx.id_to_path_map()` in its own `try/except Exception` block, returning `rejected_contest_index_unavailable` (503) on error. An empty index returns `{}` without raising, correctly returning `rejected_invalid_contest` (400).
2. **Item 2 (Must-fix):** In `src/data_olympus/tools_write.py:1817`, restored `return _park(demotion_reason)` without `"contest_declared"`, preserving `resp.proposal_text == postimage` and default operator prompts.
3. **Item 3 (Should-fix):** Added REST route tests (`test_rest_propose_memory_rejects_contest` and `test_rest_propose_edit_contest_rejected_for_absent_doc`) under `tests/test_contest_validation.py`.
4. **Item 4 (Should-fix):** Confined `try/except` strictly to `idx.id_to_path_map()`, removed the redundant `or cid in target_ids` check, and eliminated the $O(N)$ `target_ids` calculation.
5. **Nits & Changelog:** In `pending.py::list()`, keyed contest projection strictly on `intent == "contest"`. Reverted intermediate variable in `held_locks()`. Updated `CHANGELOG.md` with the deferred read-side projection note.

All 19 tests in `tests/test_contest_validation.py` pass and `ruff check .` passes with zero errors.

ajdelaguila (changes requested):
Thanks, this is a good round. Items 2, 3 and 4 are fully addressed, and item 1 you fixed at the right layer. One thing still gates merge, and it is specifically about the regression test rather than the production change.

Reviewed `5e43290` against base `main` @ `54252f18`, all checks green.

## Must fix: the SQLite regression test does not exercise the method that was broken

This is the one blocker, and it is subtle, so let me be precise about why the current test does not hold the line.

In `test_check_contest_index_unavailable`, `_MockIndex.id_to_path_map()` raises `sqlite3.OperationalError` unconditionally. `_check_contest_index` catches it and returns 503, so the assertion passes. But the real `Index.id_to_path_map()` is never called. **If someone restored the `except sqlite3.Error: return {}` that you just removed from `index.py`, this test would still pass.** Changing the mock's exception class from `RuntimeError` to `sqlite3.OperationalError` makes the test read as though it covers the repaired contract, but the mock short-circuits the very method whose behaviour changed.

The empty-index case has the same shape: `_MockIndex({})` hands back a fabricated `{}` rather than demonstrating that a real, successfully-read, zero-row `docs` table produces `{}` without raising. That distinction between "query failed" and "read fine, no rows" is the whole point of the fix, and right now nothing pins it against the real implementation.

What I am asking for: exercise the real `Index.id_to_path_map()`.

- For the 503 path, induce the failure in `execute()` or `fetchall()`, **after the connection succeeds**. A missing or corrupted `docs` table on a real index object works; so does patching at the connection boundary. The point is that the failure must travel through the real method so the test fails if the swallow comes back.
- For the 400 path, build a real index with a valid but empty `docs` table and assert `rejected_invalid_contest`.

Keep the existing mock-based tests if you like, they are useful unit coverage of `_check_contest_index`'s branching. This is an additional test that guards the shared contract you changed.

## Should fix

**The new contract is undocumented.** `id_to_path_map`'s docstring still describes only what it returns, with no mention that it now propagates `sqlite3.Error`. Its sibling methods in the same class still swallow `sqlite3.Error` and return empty, so a maintainer reading the neighbours would reasonably infer the opposite contract and reintroduce the swallow. One sentence on the method saying database errors propagate and that callers must classify them would prevent that.

**The changelog does not record the behaviour change.** The entry covers duplicate `contradicts` and the deferred read-side projection, but not the `id_to_path_map` contract change. That change has an observable effect beyond contest: a failed index read is now classified as unknown rather than silently as "not in force". Worth a line, since it is the kind of thing someone debugging an outage will look for.

**Stale docstring, third instance in this PR.** The module docstring of `tests/test_contest_validation.py`, line 15, still says the proposal is demoted `with demotion_reason="contest_declared"`, which is exactly the behaviour you removed in this commit, and the test below it now correctly asserts `is None`. I am flagging the pattern rather than just the line: this is the third time in this PR that prose has outlived the code it describes. Worth a grep for the old value before the next push.

## Correcting myself on the governed-lane effect

In my own analysis I wrote that letting `sqlite3.Error` propagate closes a fail-open hole in governed-lane write protection. That overstates it, and I would rather correct it here than let it stand in the record.

The classifier improvement is real: a query failure used to produce `{}`, then no matching id, then `TARGET_NOT_IN_FORCE`, and it now produces `TARGET_UNKNOWN`, which fails closed. But there is already an independent backstop in the write path that judges the target's in-force state from its current bytes on the refreshed base and deliberately consults nothing from the index. So a SQL failure did not previously defeat governed-target protection outright. Your change repairs a fail-open in the index-based classification specifically. Still a genuine improvement, and a good reason to keep the fix, just narrower than I first described it.

## Not blocking, for the record

`derive_running_contest()` recognises `contradicts`-only metadata, so keying the `list()` projection strictly on `intent == "contest"` rests on there being no writer that produces `contradicts` without `intent`. I checked and found none, and contest persistence is new in this PR, so there are no entries predating it. That makes it a safe deployment assumption rather than a code-enforced guarantee. No action needed now; it is only a hazard if a future writer sets one without the other.

Fix the regression test and I will re-review promptly. The production code itself I am happy with.

RemanenetSpy:
@ajdelaguila All requested items addressed in commit `8d3ea55`:

1. **Real SQLite Index Regression Test (Must-fix):**
   Added `test_check_contest_index_real_sqlite_contract` in `tests/test_contest_validation.py` exercising real `Index` instances on temporary SQLite databases:
   - **503 Path:** Creates a real SQLite database file where the `docs` table is absent. `Index.id_to_path_map()` executes `SELECT id, path FROM docs`, and the unswallowed `sqlite3.OperationalError` propagates up to `_check_contest_index`, producing `rejected_contest_index_unavailable` (503). If the `except sqlite3.Error: return {}` swallow were reintroduced, `id_to_path_map()` would return `{}` and yield 400, immediately failing the test.
   - **400 Path:** Creates a real SQLite database with a valid, empty `docs` table (`CREATE TABLE docs (id TEXT PRIMARY KEY, path TEXT)`). `Index.id_to_path_map()` executes cleanly and returns `{}` without raising, producing `rejected_invalid_contest` (400) and distinguishing zero rows from operational query failure.
   - **Positive & Self-Contradiction Path:** Seeds a valid committed row (`('STD-001', 'universal/STD-001.md')`) into `docs`, verifying that a valid target returns `None` and self-contradiction returns 400.

2. **Contract Documentation (Should-fix):**
   Extended the docstring of `Index.id_to_path_map` in `src/data_olympus/index.py` with an explicit notice that `sqlite3.Error` propagates to callers, requiring callers to classify operational database errors rather than treating empty returns as zero indexed documents.

3. **Changelog Record (Should-fix):**
   Updated `CHANGELOG.md` under `[Unreleased]` documenting that `Index.id_to_path_map` now propagates `sqlite3.Error` so failed index reads fail closed as unknown/unavailable rather than misclassifying as target not in force.

4. **Prose Cleanup (Should-fix):**
   Updated line 15 in the module docstring of `tests/test_contest_validation.py` to state that high-confidence proposals park to `pending_confirmation` without auto-committing, removing the stale `"contest_declared"` reference.

All 20 tests in `tests/test_contest_validation.py` (and 31 total across `test_contest_validation.py` + `test_running_contest.py`) pass in 2.86s, and `ruff check .` passes with zero errors.

ajdelaguila:
Reviewed `8d3ea55`. The blocker from the previous round is resolved, and so are the three should-fix items. No new code findings, one wording nit below.

## Blocker: resolved

`test_check_contest_index_real_sqlite_contract` drives the real `Index.id_to_path_map()` against real SQLite files, which is what the previous test could not do. I checked that it actually pins the contract rather than merely exercising it:

* Case 1 builds a database containing only `wrong_table`, so `_connect()` succeeds and `SELECT id, path FROM docs` raises `sqlite3.OperationalError` (a `sqlite3.Error`). `Index.__init__` stores state without creating schema and `_connect()` only creates the parent directory, connects and sets the row factory, so nothing intercepts the raise. With the fix in place it propagates, `_check_contest_index` classifies it, and the assertion on `rejected_contest_index_unavailable` holds. If the swallow were restored in `index.py`, `id_to_path_map()` would return `{}`, the loop would fall through to `contradicted document not found in index`, and case 1 would fail on the status. That is the property the previous mock-based test was missing.
* Case 2 uses a valid, empty `docs` table, so the query succeeds and returns zero rows. It separates "empty index" (400) from "index unreadable" (503) on real execution rather than on a mock's behaviour.
* Case 3 covers resolution and self-contradiction against a populated table.

All three asserted reason strings match the strings `_check_contest_index` actually produces. The test asserts on `ProposeResponse.status`, not on HTTP codes; the status-to-code mapping stays covered by the REST tests added in the previous round.

I also re-checked every caller of `id_to_path_map()` now that it raises: `governed_lane.py:142`, `write_gate.py:241`, `write_gate.py:701` and your own `tools_write.py:334`. A source search finds no others. Each already wraps the call, and two improve as a direct result. `governed_lane` maps the raise to `TARGET_UNKNOWN` and fails closed, where an empty map would previously have read as `TARGET_NOT_IN_FORCE` and dropped protection. `write_gate`'s two sites deliberately fail open and keep doing so. No caller regresses.

## Should-fix items: all three addressed

* The raising contract is documented on `Index.id_to_path_map`, and it tells callers to classify the exception rather than let a failed read arrive as an empty mapping.
* CHANGELOG records the observable change.
* The stale `demotion_reason="contest_declared"` line in the test module docstring is corrected.

**Nit on the CHANGELOG wording.** "failed index reads fail closed as unknown/unavailable" is broader than what the change does. It holds for contest validation and for governed-target classification, but `write_gate.py:241` and `write_gate.py:701` both keep their deliberate empty-map fallback, so a failed read there still does not fail closed. Scoping the sentence to contest validation and governed-target classification would make it exact. Same for the docstring's "assuming an empty dictionary represents zero indexed documents": after this change an empty dictionary does reliably mean zero documents, and the thing callers must not do is turn a failure into one.

## Remaining known prerequisites, neither a code defect

**1. The branch conflicts with `main`.** 0.9.0 landed as `69e94d6` after your last push. I measured this with `git merge-tree` and then compared the three sides of each conflicted file directly, rather than inferring it:

* `src/data_olympus/pending.py`. The only overlapping region is `_acquire_lock`. `main` rewrote its body to publish the lock record atomically: write the complete record to a private temp file in the same directory, then `os.link` it onto `lock_path`, which keeps exclusive-create semantics (`os.link` raises `FileExistsError`) while removing the window in which the path existed but was empty. Your change adds the optional `intent` and `contradicts` parameters and builds a `lock_payload` dict. The resolution is to keep `main`'s temp-file-then-link sequence and add your two optional fields to the `record` dict it already builds before it is serialized, keeping your two new keyword parameters on the signature. That preserves the atomicity property the rewrite exists for, since the record is still complete before the file gets its second name.
* Your other two hunks in that file auto-merge: the lock-acquisition call site inside `PendingQueue.enqueue` and the projection inside `PendingQueue.list`. `main` also edited `enqueue`, a few lines below your hunk, to stop releasing the lock when the entry was already published; the two edits do not overlap.
* `CHANGELOG.md` conflicts only because `main` inserted a `## [0.9.0]` section directly below `## [Unreleased]`. Your entry stays under `## [Unreleased]`, above that new section.

`src/data_olympus/index.py`, `models.py`, `rest_api.py` and `server.py` all auto-merge.

**2. CI results for this head are unverified.** I see no workflow runs recorded against `8d3ea55`, where each earlier head of this branch has its full set. Resolve the conflicts, then confirm the required checks run and pass on the rebased head.

## Next step

Rebase onto current `main` and confirm the checks go green. I will re-review promptly on the new head; if the rebase is clean and nothing else moves, I have no further findings to raise, and the decision to merge is the maintainer's.

ajdelaguila:
Ran a full review of the whole change this time, not another pass over the delta, with an independent reviewer over the complete diff plus the surrounding code on current `main`. Summary first: the contest work itself is in good shape, two things in it are worth fixing, one is an open design question rather than a defect, and one finding is inherited from shared code and is mine to fix rather than yours.

## Inherited finding, maintainer-owned

A separate security finding affects shared infrastructure that this PR builds on. It is inherited rather than introduced by this PR, and I am handling it privately as maintainer. Please coordinate anything related privately.

Please do not add handling for it here. Two practical consequences for you:

- The shared fix is written and going through review now. When it lands, rebasing onto it is all this PR needs from you on that point.
- This PR should incorporate that fix before it merges. I want to be precise that this is a real sequencing dependency and not just an artifact of the current conflict: the conflict and the missing CI are going to get resolved, and once they are they provide no protection at all.

## Worth fixing in this PR

**1. The scalar form of `contradicts` is mangled on the way out.** In `PendingQueue.list()` the projection is:

```python
"contradicts": list(entry["meta"].get("contradicts") or []),
```

`derive_running_contest` in the same file documents `contradicts` as "a target identifier string **or** list of identifier strings" and normalizes both through `_as_id_list`. Your projection assumes the list form. A record carrying `{"intent": "contest", "contradicts": "STD-001"}` is listed as `["S", "T", "D", "-", "0", "0", "1"]`, and a truthy non-iterable raises during listing instead. Your own writer always produces a list, so this only bites on a record written by something else, but the two readers of the same field in the same module should not disagree. Reusing `_as_id_list` fixes both the scalar case and the raising case.

The same applies in the lock-acquisition path in `enqueue`, which requires `list`/`tuple` and silently drops a scalar rather than normalizing it.

**2. The rate-limiter ordering the tests advertise is not actually asserted.** The module docstring describes stage 1 as pre-rate-limiter and stage 2 as post, which is a real property worth protecting, but the helper tests call `_validate_contest` and `_check_contest_index` directly and the integration tests use a permissive limiter without asserting on it. Moving stage 1 after the limiter, or stage 2 before it, would leave the suite green. Driving `kb_propose_edit_fn` with a limiter and index that record their calls would pin it: invalid structure must leave both untouched, a rate rejection must prevent the index read, and a valid contest must hit the limiter before the index.

## An open design question, not a defect

`_check_contest_index` accepts any contradicted ID present in the index. That set includes drafts, expired documents, inbox entries, graph-excluded documents and reserved files.

I am not asking you to change this, because the question it raises is not one anybody has answered: **which indexed objects are legitimate contest targets?** "In force" is the obvious candidate answer and it is probably too narrow, since a retired or expired claim can still be circulating and still be worth disputing. But that argument does not obviously extend to reserved files or inbox entries either. #241 is a research issue with open questions rather than settled acceptance criteria, so there is no criterion here for your PR to have missed. I have recorded it to be settled deliberately rather than in a review thread.

## Changelog

Two corrections, both small:

- "Read-side projection in `held_locks()` is deliberately deferred pending #270 allow-list integration" is stale. #270 is merged to `main` and is in the 0.9.0 candidate, and its allow-list was written to cover exactly this case: it is a field allow-list rather than a `pending_id` special case, and its comment names "#241's contest fields" as its example. The deferral is still a reasonable choice; the stated reason for it is what no longer holds.
- The entry does not mention what is arguably the most user-visible behaviour of the feature: a valid contest always parks for confirmation, even at high confidence with auto-commit otherwise permitted. That deserves a sentence.

Also from the previous round and still open: "failed index reads fail closed as unknown/unavailable" is broader than what the change does, and "secret scanning across all string inputs" is broader than what is scanned, since strings under unknown keys and non-list `contradicts` shapes are rejected without being scanned. Neither leaks anything; both are just wider than the code.

## What the review found clean

So there is no ambiguity about the depth of this pass: no new unauthenticated disclosure, since `held_locks()` does not project the new fields and #270's allow-list covers them by construction; no weakening of the path lock, with claim and restore preserving the existing lock rather than recreating it, which your on-disk test asserts properly; stage ordering correct; duplicates and bounds enforced; `None` meaning no contest, with no falsy shape bypassing the validator; index failures and unknown IDs not echoing submitted values; the SQLite regression substantive; and no tautological assertions.

## Still needed before this can merge

The branch conflicts with `main`, and no CI has run on this head. The conflict resolution I described previously still applies, and the reviewer independently flagged the same constraint: keep `main`'s complete-record-then-`os.link` publication and its published-entry check in `enqueue`, and extend the record rather than reinstating the older lock body, or the atomicity fix and the lock-release fix would both be undone by the rebase.

So: the two items above, then a rebase once the shared fix has landed, then green checks. I will re-review promptly at each step. The merge decision is the maintainer's.

ajdelaguila:
@RemanenetSpy, a heads-up and an apology.

I am going to push the remaining work onto this branch myself rather than wait for you to do another round. I want to be straight about why, and about what that does and does not mean.

The reason is simply that I want this in the release that is being cut now, rather than the one after it. You have turned around four review rounds on this, quickly and carefully each time, and the last round answered the blocker properly: the SQLite contract test you added drives the real `Index.id_to_path_map` and genuinely pins the behaviour, which is exactly what the previous mock-based version could not do. The remaining items are small, and the thing actually holding the branch up is a conflict that only appeared because 0.9.0 landed underneath you while you were responding to my review. That is not a fault of yours and it would be a poor reason to bump your work to the next release.

So this is me removing an obstacle I put in front of you, not taking the work away from you. Your commits stay as they are, with your authorship on them. Anything I add goes on top as separate commits under mine, so the history stays honest about who wrote what.

Apologies for not waiting for you. If you would rather have finished it yourself, say so and I will stop and hand it straight back. That is a completely reasonable thing to want, and no explanation needed.

## What I am about to do

1. **Rebase onto current `main`.** Keeping `main`'s atomic lock publication (complete record to a private temp file, then `os.link`) and its `enqueue` published-entry check, with your two optional fields folded into the record it already builds and your two new keyword parameters kept on the signature. Your other hunks auto-merge.
2. **Normalize `contradicts` through `_as_id_list`**, in the `PendingQueue.list()` projection and on the lock-acquisition path, so the scalar form the `derive_running_contest` contract documents stops being mangled into single characters.
3. **Pin the rate-limiter ordering with real tests**, driving `kb_propose_edit_fn` with a limiter and index that record their calls, so the pre/post ordering your module docstring advertises cannot silently regress.
4. **Tighten the changelog wording** on the three points from my last comment.

I am not touching the design question about which indexed objects are legitimate contest targets. That one is genuinely open, it is not something your PR got wrong, and it should be settled deliberately rather than folded into a rebase.

There is also a separate maintainer-owned fix this branch needs to sit on top of, which I mentioned before and am handling. That is mine, not yours.

## Credit

The changelog entry credits you, and it stays. This slice is your design and your implementation: the two-stage validation split around the rate limiter, the decision to signal through `meta["intent"]` rather than overloading `demotion_reason`, and dropping the `held_locks()` projection until the allow-list was ready were all your calls, and all three were the right ones. The review rounds were long because the feature touches the write path and the lock, not because the work was weak.

Thanks for sticking with it. I will report back here once it is rebased and green.

RemanenetSpy:
Thanks @ajdelaguila, that makes complete sense. Fully onboard with you handling the rebase and merge for the release cut. Appreciate the collaboration throughout the review rounds!

ajdelaguila:
@RemanenetSpy, your contest work is live: it shipped today in [v0.9.0](https://github.com/knaisoma/data-olympus/releases/tag/v0.9.0).

It landed through #281, which is this branch rebased onto `main` with your five commits unchanged and still under your name, plus maintainer follow-ups on top. Closing this PR in favour of that one is purely mechanical. The work, and the credit for it, are yours, and the release notes say so: the two-stage validation around the rate limiter and the dedicated `intent` field are your design, and holding back the read-side projection until the access controls were ready was your call.

On the way to the release it was also checked against the published package and a live deployment. A malformed contest, an unknown contradicted id and a self-contradiction are each refused without anything being parked, and a valid contest parks with its dispute rationale kept separate from the operational reason, exactly as you built it.

The open question about which indexed documents are legitimate contest targets is still open on #241, which stays open for the later slices.

Thank you for sticking with this through several review rounds. It made the release better.

